PR #9, opened 2019-10-15
Introduce CPU module

dmlloyd:


dmlloyd:
Well, that's a problem.

The MR JAR support requires a JBoss fork of the `maven-compiler-plugin`.

kenfinnigan:
where's the forked version live?

do we need to add JBoss Repo? Is it ok for us to use that forked version upstream?

dmlloyd:
The forked version is at https://github.com/jboss/maven-compiler-plugin.

dmlloyd:
Maybe we could use some more voices shouting about https://issues.apache.org/jira/browse/MCOMPILER-320 to move things along.

PR #26, opened 2020-04-27
Added annotations module.

radcortez:


pom.xml
```diff
@@ -45,6 +45,7 @@
     </scm>
 
     <modules>
+        <module>annotations</module>
```
dmlloyd, line 48: Just a nitpick, but all the other modules are non-plural (constraint, expression, etc.)  They probably should be consistent.

radcortez: Yeah, I was also wondering about singular or plural. I've ended up using plural because other annotation based projects use the plural version: `jboss-annotations`, `hibernate-annotations`, etc.

annotation/src/main/java/io/smallrye/common/annotation/Experimental.java
```diff
@@ -0,0 +1,18 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that specifies that an element is experimental and may change without notice.
+ */
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.TYPE })
+public @interface Experimental {
+    String value() default "";
```
dmlloyd, line 16: I guess the value is meant to be some explanation as to why it is experimental?

dmlloyd: Also I would add `@Documented` to this.

radcortez: Yes, let me add a javadoc in there.

dmlloyd:
Looks like there's a formatting problem.

radcortez:
Yeah, changed without local build :D

annotation/src/main/java/io/smallrye/common/annotation/Experimental.java
```diff
@@ -0,0 +1,24 @@
...
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.TYPE })
+public @interface Experimental {
+    /**
+     * Describes why the annotated element is experimental.
+     * 
+     * @return the experimental description.
+     */
+    String value() default "";
```
kenfinnigan, line 23: Did you consider not providing a default value, to ensure an actual value is entered by anyone using the annotation?

jmartisk: Maybe this was already discussed (and dismissed) but what about having an enum with values like CHANGES_SPECIFICATION, NOT_COVERED_IN_SPECIFICATION, etc? It would enable us to scan the code with tooling to assess how much we've diverged from the spec. An optional String description could be kept too.

annotation/src/main/java/io/smallrye/common/annotation/Experimental.java
```diff
@@ -0,0 +1,24 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that specifies that an element is experimental and may change without notice.
+ */
+@Inherited
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.TYPE })
```
jmartisk, line 16: What about some other targets?
`FIELD` potentially makes sense I think (for a CDI producer field for example)
`PACKAGE` too if we add a whole new package somewhere (if it contains a lot of stuff, do we want to separately annotate everything that is contained in that package?)

annotation/src/main/java/io/smallrye/common/annotation/Experimental.java
radcortez: > Did you consider not providing a default value, to ensure an actual value is entered by anyone using the annotation?

I did. I thought that in some cases, the java docs would be self sufficient to provide the explanation (in case of new classes for instance), and then requiring description in Experimental might be duplicating the information.

I have no issue making it mandatory.

radcortez: > Maybe this was already discussed (and dismissed) but what about having an enum with values like CHANGES_SPECIFICATION, NOT_COVERED_IN_SPECIFICATION, etc? It would enable us to scan the code with tooling to assess how much we've diverged from the spec. An optional String description could be kept too.

Actually, we didn't thought about that. Might be a good idea. We do need to figure which values should we use. Do you want to come with a proposal? Or do you think these are enough?

annotation/src/main/java/io/smallrye/common/annotation/Experimental.java
radcortez: We can include both.

annotation/src/main/java/io/smallrye/common/annotation/Experimental.java
jmartisk: I can send a PR and we'll see what others think. I suppose that if we do this, then the description should stay optional, there's no reason to require both to be specified.

radcortez: Please, do :)

Ladicek: I did propose having an enum, but I find it hard to come up with a comprehensive set of values. IMHO, mandatory `String value()` is enough.

kenfinnigan: It was discussed in the policy document.

The outcome was to stick with something flexible for now, just a description, and see if it made sense to constrain it further as we begin to use it

annotation/src/main/java/io/smallrye/common/annotation/Experimental.java
kenfinnigan: Doesn't hurt to support all targets, even if we don't think it necessarily makes sense right now

annotation/src/main/java/io/smallrye/common/annotation/Experimental.java
kenfinnigan: > > Did you consider not providing a default value, to ensure an actual value is entered by anyone using the annotation?
> 
> I did. I thought that in some cases, the java docs would be self sufficient to provide the explanation (in case of new classes for instance), and then requiring description in Experimental might be duplicating the information.
> 
> I have no issue making it mandatory.

It would be good to force a description, otherwise it's up to the developer to actually write javadocs.

radcortez:
Addressed mandatory for value here: https://github.com/smallrye/smallrye-common/pull/33

PR #28, opened 2020-04-27
Require Java 11 to build

### LINKED ISSUE #27: Build requires JDK 9 or higher
And CI needs to be updated accordingly as well.

dmlloyd:
Fixes #27

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,6 +99,26 @@
     <build>
         <plugins>
             <plugin>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <version>3.0.0-M2</version>
+                <executions>
+                    <execution>
+                        <id>enforce-versions</id>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -98,6 +98,26 @@
...
+                <executions>
+                    <execution>
+                        <id>enforce-versions</id>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <requireJavaVersion>
+                                    <version>11</version>
```
kenfinnigan, line 113: Any reason for the requirement?

It's still compiling against JDK 8 APIs right?

dmlloyd: It produces an MR JAR for at least one module, so you need to build with an appropriate JDK to support all the layers.  The base layer is Java 8 though.

Ladicek:
The produced artifacts are still Java 8 compatible?

Ladicek:
OK, I see the question was already answered in the comment. Thanks :-)

PR #32, opened 2020-04-28
Introduce reason for being experimental

jmartisk:
Per https://github.com/smallrye/smallrye-common/pull/26#discussion_r416415065
Let's see what people think about this


jmartisk:
Looks rejected for the time being! I'll close this and we might revisit in the future

PR #36, opened 2020-07-15
Introduce smallrye-common-io

gastaldi:


radcortez:
Does it make sense to add some tests?

gastaldi:
I thought about that too, but I wasn't certain what would that do since all methods just do `new JarFile`

gastaldi:
That would be nice, thanks :+1:

PR #37, opened 2020-07-15
Introduce JarEntries, tests and io.smallrye.common.io.jar package

gastaldi:
JarEntry.getRealName() was introduced in JDK 10 only.

Necessary for https://github.com/quarkusio/quarkus/issues/10756

### DIFF
--- a/io/pom.xml
+++ b/io/pom.xml
@@ -21,8 +21,8 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/io/pom.xml
+++ b/io/pom.xml
@@ -26,6 +26,23 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <version>3.0.0-M5</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>integration-test</goal>
+                            <goal>verify</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
     <profiles>
         <profile>
             <id>coverage</id>
--- a/io/pom.xml
+++ b/io/pom.xml
@@ -13,6 +13,19 @@
     <artifactId>smallrye-common-io</artifactId>
     <name>SmallRye Common: IO</name>
 
+    <dependencies>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-depchain</artifactId>
+            <type>pom</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
     <profiles>
         <profile>
             <id>coverage</id>
--- a/io/src/main/java/io/smallrye/common/io/JarEntries.java
+++ b/io/src/main/java/io/smallrye/common/io/JarEntries.java
@@ -1,0 +1,13 @@
+package io.smallrye.common.io;
+
+import java.util.jar.JarEntry;
+
+public class JarEntries {
+    /**
+     * Returns the real name of this {@link JarEntry}. On Java 8, it returns the {@link JarEntry#getName()}
+     * On Java 10+, a getRealName() method was added
+     */
+    public static String getRealName(JarEntry jarEntry) {
+        return jarEntry.getName();
+    }
+}
--- a/io/src/test/java/io/smallrye/common/io/jar/JarEntriesIT.java
+++ b/io/src/test/java/io/smallrye/common/io/jar/JarEntriesIT.java
@@ -1,23 +1,36 @@
 package io.smallrye.common.io.jar;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import java.io.File;
 import java.io.IOException;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnJre;
+import org.junit.jupiter.api.condition.EnabledOnJre;
+import org.junit.jupiter.api.condition.JRE;
 
 // This needs to be run as an integration-test
 public class JarEntriesIT {
 
     @Test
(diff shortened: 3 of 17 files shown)


io/src/main/java/io/smallrye/common/io/JarEntries.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.jar.JarEntry;
+
+public class JarEntries {
+    /**
+     * Returns the real name of this {@link JarEntry}. On Java 8, it returns the {@link JarEntry#getName()}
+     * On Java 10+, a getRealName() method was added
+     */
+    public static String getRealName(JarEntry jarEntry) {
+        return jarEntry.getName();
```
dmlloyd, line 11: What does this do on Java 9?  Just return a wrong answer?

gastaldi: Yes, unfortunately. They realized that was missing and fixed in Java 10

kenfinnigan: Do we want Java 9 specific bits?

When we move to 11, won't it be not required?

gastaldi: I have no idea when Quarkus will move to 11, but at least this class keeps consistency until that happens

kenfinnigan: But do we expect people to use JDK 9?

Or is it JDK 8 or 11+?

gastaldi: Considering the [LTS releases](https://www.oracle.com/java/technologies/java-se-support-roadmap.html), I think it's 8 or 11+. I don't see JDK 9 as an alternative now, but that's just my opinion :wink:

dmlloyd: Don't forget that this is no Quarkus. I don't think it would do harm to support Java 9 if we can figure out a way to do so.

gastaldi: The only way I see is using reflection (to invoke package protected methods in `JarEntry`) . Ugly but may do the job for the JDK9 users, WDYT?

Ladicek: I think we don't really have to care about Java 9 and 10.

dmlloyd: Then we should somehow expressly prevent the library from running on 9 or 10, IMO.  I think focusing only on LTS probably isn't a good idea and sets a bad precedent, encouraging bit-rot for no good reason (i.e. avoiding 5 minutes of work).

gastaldi: @Ladicek makes a good point. The project is already built in JDK 11, why do you we need to care about 9 or 10?

gastaldi: Ah but `maven.compiler.target` is set to 8, ignore me :stuck_out_tongue:

Ladicek: I'm not advocating LTS-only focus. At the same time, I do think it doesn't make much sense to care about non-LTS releases that are older than some existing LTS.

However, if it really is 5 minutes of work, then sure, that's better than defining an explicit support policy :-)

gastaldi:
There you go, now it should work with JDK 9+

io/pom.xml
```diff
@@ -13,6 +13,35 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <version>3.0.0-M5</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>integration-test</goal>
```
Ladicek, line 38: Missing `<goal>verify</goal>`?

dmlloyd:
Also might need to update the CI config... let me look into that.

io/src/test/java/io/smallrye/common/io/jar/JarEntriesIT.java
```diff
@@ -0,0 +1,23 @@
...
+
+// This needs to be run as an integration-test
+public class JarEntriesIT {
+
+    @Test
+    public void shouldUseRealName() throws IOException {
+        File tmpFile = MultiReleaseJarGenerator.generateMultiReleaseJar();
+        JarFile jarFile = JarFiles.create(tmpFile);
+        JarEntry jarEntry = jarFile.getJarEntry("foo.txt");
+        assertEquals("META-INF/versions/9/foo.txt", JarEntries.getRealName(jarEntry));
```
Ladicek, line 20: Does this rely on being run on Java 9+? (Not sure if that would be bad or not.)

io/src/test/java/io/smallrye/common/io/jar/JarFilesTest.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.junit.Test;
+
+public class JarFilesTest {
+
+    @Test
+    public void shouldReadMultiReleaseJars() throws IOException {
+        File tmpFile = MultiReleaseJarGenerator.generateMultiReleaseJar();
+        JarFile jarFile = JarFiles.create(tmpFile);
+        assertTrue(jarFile.isMultiRelease());
```
Ladicek, line 17: Does this rely on being run on Java 9+? (Not sure if that would be bad or not.)

io/src/test/java/io/smallrye/common/io/jar/JarEntriesIT.java
gastaldi: Ah yes, I need to enable it only for JDK9+ envs. Good catch

dmlloyd:
See #38.

io/src/test/java/io/smallrye/common/io/jar/JarFilesTest.java
gastaldi: Yes, I should probably enable it for JDK9+ builds too. Thanks!

dmlloyd: I think it would, which will cause the Java 8 tests to fail.  It might be worth skipping associated tests in that case.

dmlloyd: Here's an idea: add a JDK-specific `isMultiRelease(JarFile)` method.  You could even make Java 8 recognize them by browsing the manifest.

dmlloyd:
A rebase is needed now (sorry about that!)

dmlloyd:
It looks like the tests only ran with java8 and default (java 11).  Looking into it...

dmlloyd:
The tests run when I locally check out this PR, so there must have been a problem with #38.  I'll look into that separately.  AFAICT this PR can be merged (just pending @kenfinnigan).

kenfinnigan:
No need to pend on me, I was just sticking my nose in yesterday

dmlloyd:
OK :+1:

gastaldi:
Yay! Thank you! Can I haz a release soon?

radcortez:
Yes. We can push one now.

radcortez:
Well we need to fix the tests first :)

PR #38, opened 2020-07-16
Update build.yml to support JDK 9/10 testing

dmlloyd:
Due to the `smallrye-common-io` proposed changes (#36 / #37)


dmlloyd:
Just checked over the logs to ensure that the change took effect, and it looks OK.

PR #40, opened 2020-07-16
Use AdoptOpenJDK's Java setup

dmlloyd:
This allows us to have a consistent & functional env var for the JDK installation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,28 +25,31 @@ jobs:
       - uses: actions/checkout@v2
         name: checkout
 
-      - uses: actions/setup-java@v1.3.0
+      - uses: AdoptOpenJDK/setup-java@v1
         name: set up JDK 11
         with:
-          java-version: 11
+          version: 11
 
-      - uses: actions/setup-java@v1.3.0
+      - uses: AdoptOpenJDK/setup-java@v1
         name: set up JDK 10
         with:
-          java-version: 10
+          version: 10
+          targets: 'JAVA_HOME_10'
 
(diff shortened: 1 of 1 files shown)


radcortez:
Hum, maybe you want to use the one from @cescoffier :)

https://github.com/marketplace/actions/setup-java-using-binaries-from-adoptopenjdk

cescoffier:
Nope, you want the one provided by adoptOpenJdk.

dmlloyd:
We need to be able to specify an env var for the JDK release, and setup-java and most of its derivatives don't support that.

dmlloyd:
I proposed fix for the Jacoco problem in smallrye/smallrye-parent#148.

cescoffier:
@dmlloyd you can use a matrix as I did for mutiny:
https://github.com/smallrye/smallrye-mutiny/blob/master/.github/workflows/build-master.yml#L17

radcortez:
Lets fix Parent, release and fix this one.

dmlloyd:
> @dmlloyd you can use a matrix as I did for mutiny:
> https://github.com/smallrye/smallrye-mutiny/blob/master/.github/workflows/build-master.yml#L17

In this case we explicitly do *not* want a matrix.  The POM already knows how to test with each JDK, and using a matrix will dramatically increase the amount of time it takes to build the project; anyway, to properly use the SmallRye MR JAR support, the POM must only be run by the latest supported JDK.

dmlloyd:
I guess then this is waiting for smallrye/smallrye-parent#149 to get merged so we can do a release of that project, bump up the version here, and get a green board for merging.

PR #44, opened 2020-07-20
Add formatter cache directory to .gitignore

dmlloyd:


dmlloyd:
I think that part would have to be done on smallrye-parent.  I'll check into it.

dmlloyd:
I'm merging manually, because the build will never run due to the change being made to an ignored file.

PR #45, opened 2020-07-22
Support defining classes in multiple JDK versions.

radcortez:
This is just a proposal to start the discussion on how we want to support this. The main motivator is the mapping feature in SmallRye Config:

https://github.com/radcortez/smallrye-config/commit/b5dea58d990ab71691f955f4b5fa20810fd6f037#r40821102

### DIFF
--- a/classloader/pom.xml
+++ b/classloader/pom.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.smallrye.common</groupId>
+        <artifactId>smallrye-common-parent</artifactId>
+        <version>1.1.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>smallrye-common-classloader</artifactId>
+    <name>SmallRye Common: Classloader</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <profiles>
+        <profile>
+            <id>coverage</id>
+            <properties>
+                <argLine>@{jacocoArgLine}</argLine>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.jacoco</groupId>
+                        <artifactId>jacoco-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>report</id>
+                                <phase>verify</phase>
+                                <goals>
+                                    <goal>report</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
+</project>
--- a/classloader/src/main/java/io/smallrye/common/classloader/ClassDefiner.java
+++ b/classloader/src/main/java/io/smallrye/common/classloader/ClassDefiner.java
@@ -1,0 +1,46 @@
+package io.smallrye.common.classloader;
+
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Field;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+import sun.misc.Unsafe;
+
+public class ClassDefiner {
+    private static final Unsafe unsafe;
+
+    static {
+        unsafe = AccessController.doPrivileged(new PrivilegedAction<Unsafe>() {
+            public Unsafe run() {
+                try {
+                    final Field field = Unsafe.class.getDeclaredField("theUnsafe");
+                    field.setAccessible(true);
+                    return (Unsafe) field.get(null);
+                } catch (IllegalAccessException e) {
+                    throw new IllegalAccessError(e.getMessage());
+                } catch (NoSuchFieldException e) {
+                    throw new NoSuchFieldError(e.getMessage());
+                }
+            }
+        });
+    }
+
+    public static Class<?> defineClass(MethodHandles.Lookup lookup, Class<?> parent, String className, byte[] classBytes) {
+        SecurityManager sm = System.getSecurityManager();
+        if (sm != null) {
+            sm.checkPermission(DefineClassPermission.getInstance());
+        }
+
+        // The className for Unsafe.defineClass must match the one in classBytes, so we can try to verify the package
+        // with the parent
+        String parentPkg = parent.getPackage().getName();
+        String classPkg = className.substring(0, className.lastIndexOf('.'));
+
+        if (!parentPkg.equals(classPkg)) {
+            throw new IllegalArgumentException("Class not in same package as lookup class");
+        }
+
+        return unsafe.defineClass(className, classBytes, 0, classBytes.length, parent.getClassLoader(), null);
+    }
+}
--- a/classloader/src/main/java/io/smallrye/common/classloader/DefineClassPermission.java
+++ b/classloader/src/main/java/io/smallrye/common/classloader/DefineClassPermission.java
@@ -1,0 +1,44 @@
+package io.smallrye.common.classloader;
+
+import java.security.Permission;
+
+public class DefineClassPermission extends Permission {
+    private static final long serialVersionUID = 142067672163413424L;
+    private static final DefineClassPermission INSTANCE = new DefineClassPermission();
+
+    public DefineClassPermission() {
+        super("");
+    }
+
+    public DefineClassPermission(final String name, final String actions) {
+        this();
+    }
+
+    public static DefineClassPermission getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public boolean implies(final Permission permission) {
+        return permission != null && permission.getClass() == this.getClass();
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        if (obj == this) {
+            return true;
+        } else {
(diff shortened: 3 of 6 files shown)


io/src/main/java9/io/smallrye/common/io/classloader/ClassDefiner.java
```diff
@@ -0,0 +1,14 @@
+package io.smallrye.common.io.classloader;
+
+import java.lang.invoke.MethodHandles;
+
+public class ClassDefiner {
+    public static Class<?> defineClass(Class<?> parent, String className, byte[] classBytes) {
+        try {
+            MethodHandles.Lookup lookup = MethodHandles.privateLookupIn(parent, MethodHandles.lookup());
```
dmlloyd, line 8: This won't work unless this exact class has access to the target parent class.  You'd have to pass in a trusted Lookup from the caller AFAICT otherwise you risk getting access exceptions in various situations.

io/pom.xml
```diff
@@ -21,7 +21,10 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
-            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm</artifactId>
```
dmlloyd, line 27: This should be test scope, shouldn't it?

io/src/main/java9/io/smallrye/common/io/classloader/ClassDefiner.java
radcortez: Hey @dmlloyd was about to message you about this PR :)

I didn't invest too much in this, or fully tested it. I just wanted to check with you if you agree with the general strategy on how to support the `defineClass`.

radcortez: Meaning if you agree with the MR Jar and have different impls for a `defineClass`.

io/pom.xml
radcortez: Sure.

io/src/main/java9/io/smallrye/common/io/classloader/ClassDefiner.java
radcortez: If the strategy is fine, I'll clean up, fix and add more tests.

dmlloyd: I agree with the general approach.  Maybe it doesn't belong in "io" though because it's not I/O?  It might be worth its own module specific to class loading.

radcortez: Yes, I thought about this as well, but I wasn't sure if it was worth to add its own module. I'll move it. Thanks!

classloader/src/main/java/io/smallrye/common/classloader/ClassDefiner.java
```diff
@@ -0,0 +1,64 @@
...
+                    throw new IllegalAccessError(e.getMessage());
+                } catch (NoSuchFieldException e) {
+                    throw new NoSuchFieldError(e.getMessage());
+                } catch (ClassNotFoundException e) {
+                    throw new NoClassDefFoundError(e.getMessage());
+                }
+            }
+        });
+
+        defineClass = AccessController.doPrivileged(new PrivilegedAction<Method>() {
```
dmlloyd, line 33: Why are we using reflection to call this method instead of calling it directly?

radcortez: The method was removed in JDK 11.

dmlloyd: But we build an MR JAR, and we build with `release=8`, so we don't have to care about that - but you *do* have to add the [`build-include-jdk-misc` file](https://github.com/smallrye/smallrye-parent#mr-jar-sun-misc) in order to make the build work.

radcortez: Let me check.

gastaldi: +1 to create a MR JAR

classloader/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.smallrye.common</groupId>
+        <artifactId>smallrye-common-parent</artifactId>
+        <version>1.1.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>smallrye-common-classloader</artifactId>
```
gastaldi, line 11: WDYT about calling this `smallrye-common-lang` since it deals with `java.lang.*` classes?

dmlloyd: So far (dating back to WildFly Common, this project's predecessor) the naming convention has been based on the functionality of the classes, not on any particular correspondence to a JDK package.  I think this is less confusing, particularly when some functionality may abstractly cover more than one JDK package.

classloader/src/main/java/io/smallrye/common/classloader/ClassDefiner.java
radcortez: Ok done. I've added a custom fix here, because parent had a bug in the scope: https://github.com/smallrye/smallrye-parent/pull/156

classloader/src/main/java/io/smallrye/common/classloader/ClassDefiner.java
```diff
@@ -0,0 +1,41 @@
...
+                    throw new IllegalAccessError(e.getMessage());
+                } catch (NoSuchFieldException e) {
+                    throw new NoSuchFieldError(e.getMessage());
+                }
+            }
+        });
+    }
+
+    public static Class<?> defineClass(MethodHandles.Lookup lookup, Class<?> parent, String className, byte[] classBytes) {
+        // The className for Unsafe.defineClass must match the one in classBytes, so we can try to verify the package
```
dmlloyd, line 30: The possibility exists that this code could be consumed from WildFly or other environments that require security manager support.  Since this operation is implicitly privileged, we need to add a permission check here.

classloader/src/main/java9/io/smallrye/common/classloader/ClassDefiner.java
```diff
@@ -0,0 +1,21 @@
+package io.smallrye.common.classloader;
+
+import java.lang.invoke.MethodHandles;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+public class ClassDefiner {
+    public static Class<?> defineClass(MethodHandles.Lookup lookup, Class<?> parent, String className, byte[] classBytes) {
+        return AccessController.doPrivileged(new PrivilegedAction<Class<?>>() {
```
dmlloyd, line 9: The permission check must also happen here.

classloader/src/main/java/io/smallrye/common/classloader/ClassDefiner.java
radcortez: something like `new RuntimePermission("defineClass")` or do you prefer another name?

dmlloyd: Don't use RuntimePermission because it has a specific meaning in the JVM already.  I'd just create a `DefineClassPermission `and have it extend `Permission` with no action string supported.  That'd be the simplest & safest solution.

radcortez: Ok!

classloader/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <configuration>
+                            <skipTests>true</skipTests>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
```
dmlloyd, line 48: With the changes in the new parent, this shouldn't be needed any longer.

radcortez: Yes, but we need a release of the parent. If we can do one now, we can remove this.

classloader/src/main/java9/io/smallrye/common/classloader/ClassDefiner.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.lang.invoke.MethodHandles;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+public class ClassDefiner {
+    public static Class<?> defineClass(MethodHandles.Lookup lookup, Class<?> parent, String className, byte[] classBytes) {
+        SecurityManager sm = System.getSecurityManager();
+        if (sm != null) {
+            sm.checkPermission(new DefineClassPermission("defineClass"));
```
dmlloyd, line 11: Can you lift the permission object into a constant?

classloader/src/main/java/io/smallrye/common/classloader/DefineClassPermission.java
```diff
@@ -0,0 +1,11 @@
+package io.smallrye.common.classloader;
+
+import java.security.BasicPermission;
+
+public class DefineClassPermission extends BasicPermission {
```
dmlloyd, line 5: Don't extend `BasicPermission` because the permission class defines no actions.

radcortez: Done. Let me know if the new implementation is better :)

classloader/src/main/java/io/smallrye/common/classloader/DefineClassPermission.java
```diff
@@ -0,0 +1,44 @@
...
+        return permission != null && permission.getClass() == this.getClass();
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        if (obj == this) {
+            return true;
+        } else {
+            return obj != null && obj.getClass() == this.getClass();
+        }
```
dmlloyd, line 32:
```suggestion
        return obj instanceof DefineClassPermission;
```

dmlloyd:
One comment, otherwise everything looks OK.

radcortez:
Thanks :)

PR #46, opened 2020-07-24
Fix tests in Commons IO.

radcortez:


dmlloyd:
This change is not correct: the default test profile is intended to run tests at the build JDK level and should not normally be disabled.

radcortez:
The tests are failing with the default profile, so we probably need @gastaldi to have a look and check if there is something to be done over there.

dmlloyd:
I was looking at the MR JAR code and I think there is a better fix up at the parent level.  I'm going to work something up and see what I get.

PR #49, opened 2020-07-28
Introduce SmallRye Common Version module

dmlloyd:


version/src/main/java/io/smallrye/common/version/VersionScheme.java
```diff
@@ -0,0 +1,134 @@
...
+          | version "." underscore-sequence
+
+underscore-sequence ::= dash-sequence
+                      | underscore-sequence "_" dash-sequence
+
+dash-sequence ::= plus-sequence
+                | dash-sequence "-" plus-sequence
+
+plus-sequence ::= parts
+                | plus-sequence "+" parts
```
Ladicek, line 89: Any particular reason to write these productions as left-recursive? It seems to me that

```
plus-sequence ::= parts
                | plus-sequence "+" parts
```

is equivalent to

```
plus-sequence ::= parts ("+" parts)*
```

And you already use `+` elsewhere in the grammar to mean repetition (with at least 1 occurence). So using `*` (to mean repetition, no occurrence required) should be fine IMHO -- both chars are quite conventional in grammars.

dmlloyd: No reason; I just like left-recursive grammars. :)

PR #53, opened 2020-07-31
Introduce SmallRye Common BOM

dmlloyd:


gastaldi:
I think this is safe to be merged and released?

dmlloyd:
I'll let @radcortez decide if we should release again right away.

radcortez:
Should be fine. I'll do it later.

PR #57, opened 2020-08-27
initial import of power annotations

t1:
As discussed with @phillip-kruger: initial import of the code from https://github.com/t1/power-annotations

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -57,6 +57,11 @@
             </dependency>
             <dependency>
                 <groupId>io.smallrye.common</groupId>
+                <artifactId>power-annotations-api</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.smallrye.common</groupId>
                 <artifactId>smallrye-common-version</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -55,8 +55,9 @@
         <module>cpu</module>
         <module>expression</module>
         <module>function</module>
-        <module>net</module>
         <module>io</module>
+        <module>net</module>
+        <module>power-annotations</module>
         <module>version</module>
     </modules>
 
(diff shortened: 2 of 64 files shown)


dmlloyd:
I'm not so sure that this belongs in the Common project.  I don't know that it passes the "common code" test; rather, this seems to implement a solution within a space.

t1:
@dmlloyd & @Ladicek: this was my first thought, too. But the perspective of this tool is to be used everywhere... the more places it's used in, the better it gets; and the higher the pressure grows on everything else to use it, too.

In the JEE/MP world, nobody wants to depend on a single implementation, so the api (and tck and the to-be-developed spec) is going to move somewhere else... maybe MP, maybe JEE, maybe Apache. Then anybody will be free to implement it on their own.

dmlloyd:
Sure, but consider that in this case you're not really providing commonly accepted annotations; rather, it's more of a push towards a specific style or methodology (whatever the merits and virtues of that may be).  I think in this case, having a separate project is warranted.  SmallRye Common is definitely *not* intended as a carrier project to drive adoption of other things using a "fait accompli" sort of strategy like you suggest.  The more opinionated this project gets, the farther it strays from its mission, and the less likely it is to be adopted as a result.

t1:
I didn't suggest to add it to smallrye commons! I totally agree to your point. I'm only speculating about the reasons why @radcortez and @phillip-kruger may have agreed on this solution.

Ladicek:
I totally agree with David that this shouldn't be put into SmallRye Common, because it clearly is not a "common" (widely accepted, agreed on, used, etc.) code. A separate project in SmallRye makes sense to me, but I personally don't see why it can't be a personal project or a dedicated GitHub org. It's not like getting this into SmallRye brings it any closer to becoming "common" :-)

Also, I'd suggest perhaps starting a discussion on the SmallRye mailing list, to get more SmallRye people involved. (Similarly to how David started a discussion on his YAML -> `module-info.class` generator, which I have seemingly shot down :-) Sorry David!)

dmlloyd:
> I totally agree with David that this shouldn't be put into SmallRye Common, because it clearly is not a "common" (widely accepted, agreed on, used, etc.) code. A separate project in SmallRye makes sense to me, but I personally don't see why it can't be a personal project or a dedicated GitHub org. It's not like getting this into SmallRye brings it any closer to becoming "common" :-)

Okay, let's proceed along those lines.  I'll close this PR.

> Also, I'd suggest perhaps starting a discussion on the SmallRye mailing list, to get more SmallRye people involved. (Similarly to how David started a discussion on his YAML -> `module-info.class` generator, which I have seemingly shot down :-) Sorry David!)

Hey, I haven't given up yet.  I'm confident that you'll be back with your hat in your hand.  Just wait and see... :grin:

PR #63, opened 2020-09-29
Add ClassPathUtils, copied from Quarkus.

radcortez:


classloader/pom.xml
```diff
@@ -21,6 +21,14 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-impl-base</artifactId>
+        </dependency>
```
Ladicek, line 31: Aren't they both `<scope>test</scope>`? I'm pretty sure we wouldn't want all consumers to also transitively get ShrinkWrap.

radcortez: Right :) Thanks!

Ladicek (approved):
I vaguely recall someone saying the class shouldn't have a `*Util[s]` name, but I don't remember the specifics. LGTM anyway :-)

radcortez:
> I vaguely recall someone saying the class shouldn't have a `*Util[s]` name, but I don't remember the specifics. LGTM anyway :-)

Yes, I also heard that before, but I'm terrible for names. Also, wanted to keep the same name as Quarkus :)

Thanks!

PR #65, opened 2020-10-19
Add the @Blocking and @NonBlocking annotations

cescoffier:
Follow-up on the discussion from https://github.com/quarkusio/quarkus/pull/12418.

Add the `@Blocking` and `@NonBlocking` annotation.

@Ladicek Can you check and comment? 
@mkouba I dropped the Qualifier part for now (hopefully for good). I will see how I can work around this in Quarkus-gRPC (@michalszynkiewicz  FYI).

### DIFF
--- a/annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
+++ b/annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
@@ -1,0 +1,34 @@
+package io.smallrye.common.annotation;
+
+import static java.lang.annotation.ElementType.METHOD;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to indicate that the annotated method is inherently blocking and so should not be
+ * executed on a <em>non-blockable</em> thread (I/O thread, event loops...).
+ * <p>
+ * Frameworks can add support for this annotation and offload the work to another thread if the current thread cannot
+ * be blocked. It's particularly useful for frameworks using a reactive execution model. Framework relying on this
+ * annotation must specify the exact behavior:
+ * <ul>
+ * <li><em>what</em> thread is considered <em>non-blockable</em>;</li>
+ * <li>on which thread is the execution offloaded;</li>
+ * <li>whether, when the current thread can block, the execution of the annotated method is still offloaded to
+ * another thread, or stays on the same thread;</li>
+ * <li>if the execution of the method is offloaded, whether the initial thread is restored after the method
+ * execution.</li>
+ * </ul>
+ * <p>
+ * This annotation is not <em>inheritable</em>, so the user must repeat the annotation when overriding the method.
+ *
+ * @see NonBlocking
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(METHOD)
+public @interface Blocking {
+
+    // Just a marker annotation.
+}
--- a/annotation/src/main/java/io/smallrye/common/annotation/NonBlocking.java
+++ b/annotation/src/main/java/io/smallrye/common/annotation/NonBlocking.java
@@ -1,0 +1,26 @@
+package io.smallrye.common.annotation;
+
+import static java.lang.annotation.ElementType.METHOD;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to indicate that the annotated method is inherently non-blocking and so can be
+ * executed on a <em>non-blockable</em> thread (I/O threads, event loops...) without the need to offload the work to
+ * another thread. If the caller thread can be blocked, it should also be safe to execute the method on that thread.
+ * <p>
+ * It's up to the framework relying on this annotation do define the exact behavior, like <em>what</em> thread is
+ * considered as a <em>non-blockable</em> thread.
+ * <p>
+ * This annotation is not <em>inheritable</em>, so the user must repeat the annotation when overriding the method.
+ *
+ * @see Blocking
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(METHOD)
+public @interface NonBlocking {
+
+    // Just a marker annotation.
+}


radcortez (dismissed):

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
```diff
@@ -0,0 +1,24 @@
...
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to indicate that the annotated element (method or class) is inherently blocking and so should not be
```
Ladicek, line 11: What's the meaning when a _class_ is annotated? As if all relevant methods were annotated?

Ladicek: And does this class annotation apply to subclasses? (IMHO, annotating classes as a shortcut to annotating all relevant methods opens a nasty can of worms. See the Jakarta Interceptors specification for an example :-) )

mkouba: > And does this class annotation apply to subclasses?

Well, it's not annotated with `@Inherited` so it should not apply to subclasses if you follow the general common annotations rules ;-)

mkouba: Maybe we should make it `@Target({ METHOD })` only?

Ladicek: > Well, it's not annotated with `@Inherited`

Right, I missed that, sorry about the noise :-)

Ladicek: Anyway -- my main point is that the javadoc should answer these questions (or at least those that make sense :-) ).

cescoffier: I don't believe we should support inherited classes. That's just too tricky. I will add a line about that. 

About `TYPE`, it's (again) driven by the gRPC use case, where you can only intercept the whole service and not a single method. The initial model we had in mind was:

```java
@Singleton 
@Blocking 
public class MyService extends GrpcGreeterBaseImpl {

}
```

And this was resulting in registering the _blocking server call_ interceptor for the whole service (so all methods). 

A method based `@Blocking` would need a smarter interceptor that only wraps the call to blocking methods. But in the interceptor class, I don't know if the method is blocking as I don't have access to it. UPDATE: I may have found a way...

cescoffier:
Converting the PR to a draft - there is a few investigations to do before merging.

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
Ladicek: If it's a CDI interceptor, then the `InvocationContext` gives you the intercepted method :-) But I can see how the "command" pattern might require annotating a class (which implements the command).

cescoffier: Nope, it's a gRPC Server Interceptor.

Ladicek: Ah OK. This essentially means that the annotation meaning is heavily context-dependent. I should have figured that out sooner, sorry, it's Monday :-)

For example, in case of SmallRye Fault Tolerance, where I'd like to augment MicroProfile Fault Tolerance's `@Asynchronous` (which should be called `@ThreadOffload` IMHO :-) ) with `@Blocking` or `@Nonblocking` [1], I can say that the CDI interceptor treats class-level annotations like a shortcut for annotating all relevant methods.

And for gRPC, it would be something else probably.

[1] I can see 2 possible meanings: 1. `@Nonblocking` means no thread offload is really necessary and `@Blocking` is superfluous, 2. `@Nonblocking` means offload to the event loop and `@Blocking` means offload to a worker thread pool.

cescoffier: Yes, it is context-dependent, and it will be up to the "integrator" to clearly explain what it does in the scope of what the general documentation (the JavaDoc written here) says. I should actually mention that. 

Note that now you mentioned it, I wonder if I could use a CDI interceptor instead of a gRPC server interceptor. That would simplify A LOT my "try to discover which method/service is called."

dmlloyd (dismissed):

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
cescoffier: So, it cannot be a CDI interceptor (because a method can consume and return streams which need specific handling), but not a problem... I've found a way.

* the annotations don't need to a qualifiers (woohoo, I made @mkouba happy for once!)
* the annotations should be `METHOD` only.

cescoffier:
Ok, just pushed an update clarifying the usage of the annotations.

@Ladicek @mkouba can you have another look.
@FroMage it might be useful for your current project too.

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
Ladicek: Yay! :-)

annotation/src/main/java/io/smallrye/common/annotation/NonBlocking.java
```diff
@@ -0,0 +1,26 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to indicate that the annotated method is inherently non-blocking and so can be
+ * executed on a <em>non-blockable</em> thread (I/O threads, event loops...) without the need to offload the work to
+ * another thread. If the caller thread can be blocked, it should also be safe to execute the method on that thread.
+ *
+ * It's up to the framework relying on this annotation do define the exact behavior, like <em>what</em> thread is
+ * considered as a <em>non-blockable</em> thread.
+ *
```
Ladicek, line 16: You probably want to separate the paragraphs with `<p>`, not just empty line.

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
```diff
@@ -0,0 +1,34 @@
...
+ * Annotation used to indicate that the annotated method is inherently blocking and so should not be
+ * executed on a <em>non-blockable</em> thread (I/O thread, event loops...).
+ * <p>
+ * Frameworks can add support for this annotation and offload the work to another thread if the current thread cannot
+ * be blocked. It's particularly useful for frameworks using a reactive execution model. Framework relying on this
+ * annotation must specify the exact behavior:
+ * <ul>
+ * <li><em>What</em> thread is considered <em>non-blockable</em>;</li>
+ * <li>On which thread is the execution offloaded;</li>
+ * <li>Whether, when the current thread can be block, the execution of the annotated method is still offloaded to
```
Ladicek, line 19:
```suggestion
 * <li>Whether, when the current thread can be blocked, the execution of the annotated method is still offloaded to
```

or

```suggestion
 * <li>Whether, when the current thread can block, the execution of the annotated method is still offloaded to
```

?

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
```diff
@@ -0,0 +1,34 @@
...
+ * another thread, or stay on the same thread;</li>
```
Ladicek, line 20:
```suggestion
 * another thread, or stays on the same thread;</li>
```

?

annotation/src/main/java/io/smallrye/common/annotation/NonBlocking.java
cescoffier: Good catch!

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
```diff
@@ -0,0 +1,34 @@
...
+ * <li>If the execution of the method is offloaded, whether the initial thread is restored after the method
+ * execution.</li>
```
Ladicek, line 22: Nitpicking: I don't know how this works in English, but when I write bullet lists as a sentence (where the individual items end with comma or semicolon and the last one ends with a period), the items typically don't begin with an upper-case letter :-) That might be wrong -- I'd love to learn how it really should be.

Ladicek:
Added just a few minor comments, LGTM in general.

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
cescoffier: I don't know either, so I removed the uppercase.

cescoffier:
@Ladicek I've addressed your comments (Thanks!) and added you as a co-author.

Ladicek (approved):

Ladicek:
I don't really feel like a co-author, but thanks! :-)

annotation/src/main/java/io/smallrye/common/annotation/Blocking.java
```diff
@@ -0,0 +1,34 @@
...
+ * <ul>
+ * <li><em>what</em> thread is considered <em>non-blockable</em>;</li>
+ * <li>on which thread is the execution offloaded;</li>
+ * <li>whether, when the current thread can block, the execution of the annotated method is still offloaded to
+ * another thread, or stays on the same thread;</li>
+ * <li>if the execution of the method is offloaded, whether the initial thread is restored after the method
+ * execution.</li>
+ * </ul>
+ * <p>
+ * This annotation is not <em>inheritable</em>, so the user must repeat the annotation when overloading the method.
```
mkouba, line 25: Hm, isn't "overloading" a different concept? You're probably talking about "overridden" methods.

Ladicek: That's right, I totally missed that. Should be "when overriding the method".

cescoffier: You're right! I should not work on 2 PRs in parallel... (in GRPC methods cannot be overloaded :-)

cescoffier: Fixed!

Ladicek (approved):

cescoffier:
@radcortez anything else required for this PR and getting a release? Need it for Quarkus 1.10 (still a bit of time, so no rush, but I just would like to address any issues soon enough (and avoid a final half-backed rush before the code freeze)).

mkouba (approved):
Looks good.

radcortez (approved):

radcortez:
Looks good to me too. I'll merge and push a release.

cescoffier:
Thanks @radcortez !

PR #66, opened 2020-10-21
Release 1.4.1.

radcortez:


dmlloyd:
@radcortez - For future reference, any new features or APIs require a new minor version (1.5.0 in this case).  A micro release is *only* for fixes and non-API changes.

radcortez:
Ok, since they were small changes, I though they didn't require to bump to 1.5.0. Sorry about that.

We could do another release to 1.5.0. What do you think?

dmlloyd:
I don't think another release would help unless the 1.4.1 release could be deleted, but since it was already published, that cannot be done without potentially disrupting users.  It's too late to correct this one unfortunately.

PR #67, opened 2020-11-02
Allow to use relative paths when the url is a file type.

radcortez:


classloader/src/main/java/io/smallrye/common/classloader/ClassPathUtils.java
```diff
@@ -202,7 +202,7 @@ public static void consumeStream(URL url, Consumer<InputStream> consumer) throws
             }
         }
         if (FILE.equals(url.getProtocol())) {
-            try (InputStream is = Files.newInputStream(toLocalPath(url))) {
+            try (InputStream is = Files.newInputStream(Paths.get(url.getFile()))) {
```
dmlloyd, line 205: This is surprising.  I would have expected that a relative URL should have translate to a relative URI, which should have yielded a relative result from `Paths.get()`.  Where did it go wrong?

radcortez: It seems that the `Paths.get(URI)` checks if the URI is opaque and fails the result.

For instance if you try a `Paths.get(new URL("file:./a").toURI())` you get a `java.lang.IllegalArgumentException: URI is not hierarchical`. But a `Paths.get(new URL("file:./a").getFile())` works fine.

Do you have any other idea?

dmlloyd: Ah, that's because `file:./a` *is* an opaque URL, it's not a valid relative URL (see https://tools.ietf.org/html/rfc3986#section-3 for info).  A valid relative URL would be `./a` (note the lack of a "scheme" component).  Where is the `file:./a` style URL coming from?

radcortez: But due to the API check for a protocol (either FILE or JAR) it seems that the only way to have the API read a relative path is to use `new URL(file:./a)` and then just call `getFile` on it. Do you suggest to change the API?

dmlloyd: I think the API should reject invalid URLs.  Where is the URL coming from?  Any time you need to use a relative URL, the question is always "relative to what?" - typically a class path element, which will have a base URL that you should resolve against (to get an absolute URL).  I think any other usage is probably incorrect.

radcortez: > Where is the URL coming from? 

As is, it can come directly from the caller, since it is a public API, and it handles the file protocol, so nothing stops the user to send such URL with a file relative to the current working directory for instance

If we leave it as is, you just get the original exception which may be fine. With this change it may work for both cases, just for the `file` protocol :)

If you feel there is no point, I can retract the PR.

dmlloyd: OK, I assumed there was some program that was expecting this to work, and I was going to suggest fixing this program.  I'd suggest that the PR be retracted in that case.

radcortez: Sure!

PR #80, opened 2021-03-08
Introduce Process.getAllProcess

### LINKED ISSUE #79: Add method to retrieve all running processes
As discussed in https://github.com/quarkusio/quarkus/pull/15296#discussion_r589424570, it would be nice to have a method that gives you information about **all** the running processes

gastaldi:
Fixes #79

### DIFF
--- a/os/src/main/java/io/smallrye/common/os/GetAllProcessesInfoAction.java
+++ b/os/src/main/java/io/smallrye/common/os/GetAllProcessesInfoAction.java
@@ -1,0 +1,122 @@
+package io.smallrye.common.os;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.security.PrivilegedAction;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+
+final class GetAllProcessesInfoAction implements PrivilegedAction<List<ProcessInfo>> {
+
+    private final static Predicate<String> IS_NUMBER = Pattern.compile("\\d+").asPredicate();
+
+    @Override
+    public List<ProcessInfo> run() {
+        // The ProcessHandle API does not exist, so we have to rely on external processes to get this information
+        switch (OS.current()) {
+            case LINUX:
+                return readLinuxProcesses();
+            case MAC:
+                return readMacProcesses();
+            case WINDOWS:
+                return readWindowsProcesses();
+            default:
+                throw new UnsupportedOperationException(
+                        "Listing all processes is not supported in JDK 8 in " + OS.current().name());
+        }
+    }
+
+    private List<ProcessInfo> readLinuxProcesses() {
+        List<ProcessInfo> processes = new ArrayList<>();
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(Paths.get("/proc"))) {
+            for (Path procPath : stream) {
+                String name = procPath.getFileName().toString();
+                if (IS_NUMBER.test(name)) {
+                    long pid = Long.parseLong(name);
+                    try (BufferedReader reader = Files
+                            .newBufferedReader(procPath.resolve("cmdline"), StandardCharsets.UTF_8)) {
+                        String line = reader.readLine();
+                        if (line != null) {
+                            int idx = line.indexOf(0);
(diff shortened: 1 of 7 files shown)


gastaldi:
@geoand see if this change fits your use case

geoand:
Yeah that would likely work

os/src/main/java/io/smallrye/common/os/GetAllProcessesInfoAction.java
```diff
@@ -0,0 +1,12 @@
+package io.smallrye.common.os;
+
+import java.security.PrivilegedAction;
+import java.util.List;
+
+public class GetAllProcessesInfoAction implements PrivilegedAction<List<ProcessInfo>> {
```
dmlloyd, line 6:
```suggestion
final class GetAllProcessesInfoAction implements PrivilegedAction<List<ProcessInfo>> {
```

os/src/main/java/io/smallrye/common/os/GetAllProcessesInfoAction.java
```diff
@@ -0,0 +1,12 @@
...
+
+    @Override
+    public List<ProcessInfo> run() {
+        throw new UnsupportedOperationException("Listing all processes is not supported in JDK 8");
```
dmlloyd, line 10: On Linux, you could open `/proc` and iterate each directory, something like this:

```java
        try (DirectoryStream<Path> stream = Files.newDirectoryStream(Paths.get("/proc"))) {
            for (Path procPath : stream) {
                String name = procPath.getFileName().toString();
                if (name.matches("\\d+")) {
                    long pid = Long.parseLong(name);
                    try (BufferedReader reader = Files.newBufferedReader(procPath.resolve("cmdline"), StandardCharsets.UTF_8)) {
                        String line = reader.readLine();
                        int idx = line.indexOf(0);
                        String cmdLine = idx == -1 ? line : line.substring(0, idx);
                        // new ProcessInfo(pid, cmdLine);
                    }
                }
            }
        }
```

I don't know what the Mac or Windows equivalents are, but it might be worth looking into (even if it ends up just being a parse of `ps` or something).

dmlloyd (commented):
I think this idea is okay, just a couple of comments.

os/src/main/java/io/smallrye/common/os/GetAllProcessesInfoAction.java
gastaldi: @dmlloyd can you fix CI to also build in Windows and MacOS? I am developing under Fedora 33 so I don't know if the fix would work in other OSes

dmlloyd: I have a PR in testing now, just have to get Windows right and then I'll merge it.  Then you can rebase and you'll have it.

geoand: Great stuff! Thanks guys! Now I won't have to rely on on reflection to do this in Quarkus :)

dmlloyd: It turns out we have test failures on Windows.  I guess adding CI configurations was a pretty good idea. :)

dmlloyd: OK that's merged.  Go ahead and rebase and you should be tested on all three platforms.

gastaldi: Awesome! Bonus points if it's built with different JDK versions too :)

dmlloyd: This is a SmallRye MR-JAR project, so it always builds with Java 11 and tests with 8, 9, 10, and 11.

os/src/main/java/io/smallrye/common/os/GetAllProcessesInfoAction.java
```diff
@@ -0,0 +1,117 @@
...
+                    long pid = Long.parseLong(name);
+                    try (BufferedReader reader = Files
+                            .newBufferedReader(procPath.resolve("cmdline"), StandardCharsets.UTF_8)) {
+                        String line = reader.readLine();
+                        if (line != null) {
+                            int idx = line.indexOf(0);
+                            String cmdLine = idx == -1 ? line : line.substring(0, idx);
+                            processes.add(new ProcessInfo(pid, cmdLine));
+                        }
+                    }
```
dmlloyd, line 52:
```suggestion
                    } catch (IOException ignored) {
                        // ignore case where process exits right before we read cmdline
                    }
```

gastaldi:
Finally all green!

dmlloyd (approved):

PR #85, opened 2021-03-15
Add @Identifier

Ladicek:
This is an alternative to `@Named`. Unlike `@Named`, it is a proper
qualifier in CDI environment.

### DIFF
--- a/annotation/pom.xml
+++ b/annotation/pom.xml
@@ -12,6 +12,25 @@
 
   <name>SmallRye Common: Annotations</name>
 
+  <dependencies>
+    <dependency>
+      <groupId>jakarta.enterprise</groupId>
+      <artifactId>jakarta.enterprise.cdi-api</artifactId>
+      <scope>provided</scope>
+      <exclusions>
+        <!-- keep the dependency tree of this module as small as possible -->
+        <exclusion>
+          <groupId>jakarta.el</groupId>
+          <artifactId>jakarta.el-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>jakarta.interceptor</groupId>
+          <artifactId>jakarta.interceptor-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+  </dependencies>
+
   <profiles>
     <profile>
       <id>coverage</id>
--- a/annotation/src/main/java/io/smallrye/common/annotation/Identifier.java
+++ b/annotation/src/main/java/io/smallrye/common/annotation/Identifier.java
@@ -1,0 +1,52 @@
+package io.smallrye.common.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.enterprise.util.AnnotationLiteral;
+import javax.inject.Qualifier;
+
+/**
+ * String-based {@linkplain Qualifier qualifier}.
+ * Unlike {@link javax.inject.Named @Named}, this is a proper qualifier in CDI environment.
+ * <p>
+ * Identifier must always be provided.
+ * Unlike {@link javax.inject.Named @Named}, it is not deduced from the annotation use.
+ */
+@Qualifier
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD, ElementType.TYPE })
+public @interface Identifier {
+    /**
+     * The identifier.
+     *
+     * @return the identifier.
+     */
+    String value();
+
+    /**
+     * Supports inline instantiation of the {@link Identifier} qualifier.
+     */
+    final class Literal extends AnnotationLiteral<Identifier> implements Identifier {
+        private static final long serialVersionUID = 1L;
+
+        private final String value;
+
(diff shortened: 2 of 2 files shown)


Ladicek:
Context: https://github.com/smallrye/smallrye-reactive-messaging/issues/1096

CC @mkouba @cescoffier

annotation/src/main/java/io/smallrye/common/annotation/IdentifiedBy.java
```diff
@@ -0,0 +1,51 @@
...
+ * Unlike {@link javax.inject.Named @Named}, this is a proper qualifier in CDI environment.
+ * <p>
+ * Identifier must always be provided.
+ * Unlike {@link javax.inject.Named @Named}, it is not deduced from the annotation use.
+ */
+@Qualifier
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD, ElementType.TYPE})
+public @interface IdentifiedBy {
```
Ladicek, line 22: Feedback on the name would be greatly appreciated :-) See also the discussion in https://github.com/smallrye/smallrye-reactive-messaging/issues/1096

annotation/pom.xml
```diff
@@ -12,6 +12,14 @@
 
   <name>SmallRye Common: Annotations</name>
 
+  <dependencies>
+    <dependency>
+      <groupId>jakarta.enterprise</groupId>
+      <artifactId>jakarta.enterprise.cdi-api</artifactId>
+      <scope>provided</scope>
```
Ladicek, line 19: Not exactly sure about this line, but I guess we don't want to drag full CDI API with SRye Common Annotations. Any ideas?

dmlloyd: I think `provided` is appropriate for this general situation.  I'm a little concerned that this brings in a tree of dependencies though.  Could we add exclusions to keep it minimal?  Also, are we OK with inheriting the version from the SmallRye parent?

annotation/src/main/java/io/smallrye/common/annotation/IdentifiedBy.java
dmlloyd: Grammatically, if a simple-past-tense verb is what is desired in order to be consistent with `Named`, I'd recommend `IdentifiedAs` instead of `IdentifiedBy`; similarly (I think) to @mkouba, I don't think `IdentifiedBy` is perfectly clear that the value is itself the identifier.

annotation/pom.xml
Ladicek: I can prune it down to just `jakarta.inject` and `jakarta.enterprise.cdi-api` itself, though the tree of dependencies of `jakarta.enterprise.cdi-api` is rather small. But better to be explicit in this case, I guess.

Ladicek: I did this. Inheriting the CDI API version from SmallRye Parent should be fine, we really only need the `@Qualifier` annotation and the `AnnotationLiteral` superclass, both of which should be very stable.

annotation/src/main/java/io/smallrye/common/annotation/IdentifiedBy.java
Ladicek: Naming is hard! And, as a non-native speaker whose fondness of English morphology and grammar has never been terribly high, I'll just take your word for it :-)

And, I don't like "identified by" (as well as "identified as") either. If we could have a single word, that would be so much better.

dmlloyd: Fair!  Other single-word, simple-past-tense verbs with similar meaning: `Called` (seems ambiguous with invocation-related terminology), `Labelled`, `Dubbed`, or `Entitled`.  No options are particularly appealing, and several are ambiguous; CDI took the good one 🙂.

Ladicek: Yes! JPA took `@Id`, JUnit Jupiter took `@Tag`, and now we're here :-)

/me grumbles and renames `@IdentifiedBy` to `@Moniker` :rofl:

radcortez: How about `@Tagged`? Still different from JUnit `@Tag`. Or you don't want to use the same verb?

cescoffier: My concern with `Tagged` is that is leads to a lookup with is a selector matching multiple tags (à la Kubernetes). 
It can be interesting, but at the same time much more complicated.

Ladicek: I've thought about this a lot yesterday and I think I'll just rename it to `@Identifier`. It's not a verb in simple past tense, but that has never been a requirement nor a convention for qualifiers.

Ladicek: Renamed to `@Identifier`.

Ladicek:
Ready for review I guess :-)

radcortez (approved):

cescoffier (approved):

radcortez:
@Ladicek I guess you want a release ASAP? :)

Ladicek:
Ah, I actually didn't think about that, but ... yes please :-)

Issue #98, opened 2021-05-15
Provide basic NIO-based PathUtils


boris-unckel:
Current IO utility libraries suffer from their early birth and a based on a java.io and java.nio mix. 
The approach is to have the file management methods of [java.nio.file.Files](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html) with [io.smallrye.common.constraint.Assert](https://github.com/smallrye/smallrye-common/blob/main/constraint/src/main/java/io/smallrye/common/constraint/Assert.java) null checks and [java.nio.file.Paths](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Paths.html) parameters. The same methods shall be available for classic java.io.File with underlying NIO to ease legacy use. All methods shall follow a fluent style and throw java.io.UncheckedIOException in error cases.

- distinct between "exception shows result" (i.E. NotLinkException) and "exception shows error in determination" (i.E. NoSuchFileException)
- utilizing SafeDirectoryStreams (providing native filesystem support in NIO) for recursive directory deletion
- avoid NIO pitfall when deleting dead symbolic links

dmlloyd:
This PR was withdrawn a while ago. Reopen if needed.

PR #99, opened 2021-05-15
Provide basic NIO-based PathUtils (#98)

boris-unckel:
Fixes https://github.com/smallrye/smallrye-common/issues/98

### DIFF
--- a/io/src/main/java/io/smallrye/common/io/FileUtils.java
+++ b/io/src/main/java/io/smallrye/common/io/FileUtils.java
@@ -1,0 +1,574 @@
+package io.smallrye.common.io;
+
+import static io.smallrye.common.constraint.Assert.checkNotEmptyAfterTrimParam;
+import static io.smallrye.common.constraint.Assert.checkNotNullParam;
+
+import java.io.File;
+import java.io.UncheckedIOException;
+import java.net.URI;
+import java.nio.file.AccessMode;
+import java.nio.file.CopyOption;
+import java.nio.file.Files;
+import java.nio.file.LinkOption;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.FileAttribute;
+
+/**
+ * Utilities for {@link java.nio.file.Files} using classic
+ * <code>java.io.File</code> as parameter. Different to parameter overload in
+ * PathUtils the return value is File, too. Please prefer a full migration to
+ * java.nio. This is for legacy convenience only.
+ *
+ * @author Boris Unckel
+ *
+ */
+public final class FileUtils {
+
+    /**
+     * A wrapper around {@link java.nio.file.Paths#get(String, String...)} with
+     * explicit null check.
+     *
+     * @param first part of the path, must be not null, not empty and a valid path
+     * @param more optional parts of the path
+     * @return the path constructed out of first and more
+     * @throws IllegalArgumentException if first is null or empty
+     */
+    public static File get(final String first, final String... more) {
+        return Paths.get(checkNotEmptyAfterTrimParam("first", first), PathUtils.checkElementsNotEmpty(more)).toFile();
+    }
+
+    /**
+     * A wrapper around {@link java.nio.file.Paths#get(URI)} with explicit null
+     * check.
+     *
+     * @param uri for the path, must be not null and a valid path
+     * @return the path constructed out of the uri
+     * @throws IllegalArgumentException if uri is null
+     */
+    public static File get(final URI uri) {
+        return Paths.get(checkNotNullParam("uri", uri)).toFile();
+    }
+
+    /**
+     * A wrapper around
+     * {@link java.nio.file.Files#createFile(Path, FileAttribute...)} with explicit
+     * null check and unchecked {@code java.io.UncheckedIOException}.
+     *
+     * @param file of the file to create, must be not null
+     * @param attrs optional attributes of the file to create
+     * @return the created file
+     * @throws IllegalArgumentException if the file is {@code null}
+     * @throws UncheckedIOException around original IOExceptions
+     */
+    public static File createFile(final File file, final FileAttribute<?>... attrs) throws UncheckedIOException {
+        checkNotNullParam("file", file);
+        return PathUtils.doCreateFile(file.toPath(), attrs).toFile();
+    }
+
+    /**
+     * A wrapper around
+     * {@link java.nio.file.Files#createDirectory(Path, FileAttribute...)} with
+     * explicit null check and unchecked {@code java.io.UncheckedIOException}.
+     *
+     * @param dir to create, must be not null
+     * @param attrs optional attributes of the directory to create
+     * @return the created directory
+     * @throws IllegalArgumentException if the dir is {@code null}
+     * @throws UncheckedIOException around original IOExceptions
+     */
+    public static File createDirectory(final File dir, final FileAttribute<?>... attrs) throws UncheckedIOException {
+        checkNotNullParam("dir", dir);
+        return PathUtils.doCreateDirectory(dir.toPath(), attrs).toFile();
+    }
+
+    /**
+     * A wrapper around
+     * {@link java.nio.file.Files#createDirectories(Path, FileAttribute...)} with
+     * explicit null check and unchecked {@code java.io.UncheckedIOException}.
+     *
+     * @param dir to create, must be not null
+     * @param attrs optional attributes of the directory to create
+     * @return the created directory
+     * @throws IllegalArgumentException if the dir is {@code null}
+     * @throws UncheckedIOException around original IOExceptions
+     */
+    public static File createDirectories(File dir, FileAttribute<?>... attrs) throws UncheckedIOException {
+        checkNotNullParam("dir", dir);
+        return PathUtils.doCreateDirectories(dir.toPath(), attrs).toFile();
+    }
+
+    /**
+     * Create a new file if not exists or update the last modified time to now.
+     *
+     * @param file to create, must be not null
+     * @return the created or updated file
+     * @throws IllegalArgumentException if file is {@code null}
+     * @throws UncheckedIOException around original IOExceptions
+     */
+    public static File touch(final File file) throws UncheckedIOException {
+        checkNotNullParam("file", file);
+        return PathUtils.doTouch(file.toPath()).toFile();
+    }
+
+    /**
+     * A wrapper around
+     * {@link java.nio.file.Files#createSymbolicLink(Path, Path, FileAttribute...)}
+     * with explicit null check and unchecked {@code java.io.UncheckedIOException}.
+     *
+     * @param link to create, must be not null
+     * @param target to point to, must be not null
+     * @param attrs optional attributes of the link
+     * @return the created symbolic link
+     * @throws IllegalArgumentException if the link or target is {@code null}
+     * @throws UncheckedIOException around original IOExceptions
+     */
+    public static File createSymbolicLink(final File link, final File target, final FileAttribute<?>... attrs)
+            throws UncheckedIOException {
+        checkNotNullParam("link", link);
+        checkNotNullParam("target", target);
+        return PathUtils.doCreateSymbolicLink(link.toPath(), target.toPath(), attrs).toFile();
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


boris-unckel:
Hello @dmlloyd as you suggested I have prepared the NIO utils PR here in smallrye-common. Please have a look and review.

io/src/main/java/io/smallrye/common/io/FileUtils.java
```diff
@@ -0,0 +1,565 @@
...
+     * {@link java.nio.file.Files#isDirectory(Path, LinkOption...)} but with
+     * explicit null check and unchecked {@code java.io.UncheckedIOException}.
+     *
+     * @param file to be checked, must be not null
+     * @return the passed path if it is a regular file
+     * @throws IllegalArgumentException if the path is {@code null}
+     * @throws UncheckedIOException if is not a regular file or determination of
+     *         type fails
+     */
+    public static File checkIsRegularFileNoFollowLinks(final File file) throws UncheckedIOException {
```
dmlloyd, line 362: I noticed that some methods take `*Options*` flags and some do not (i.e. `CopyOptions` vs `LinkOptions`); is there a reason for the inconsistency?

io/src/main/java/io/smallrye/common/io/FileUtils.java
```diff
@@ -0,0 +1,565 @@
...
+     * @throws UncheckedIOException if directory contains elements or check
+     *         fails
+     */
+    public static File checkIsEmptyDirectory(final File dir) throws UncheckedIOException {
+        checkNotNullParam("dir", dir);
+        return PathUtils.doCheckIsEmptyDirectory(dir.toPath()).toFile();
+    }
+
+    private FileUtils() {
+        throw Assert.unreachableCode();
```
dmlloyd, line 562: It is enough to have the constructor be private IMO.  This is consistent with other classes in the project I believe.

io/src/main/java/io/smallrye/common/io/PathUtils.java
```diff
@@ -0,0 +1,2139 @@
...
+     * A wrapper around {@link java.nio.file.Paths#get(String, String...)} with
+     * explicit null check.
+     *
+     * @param first part of the path, must be not null, not empty and a valid path
+     * @param more optional parts of the path
+     * @return the path constructed out of first and more
+     * @throws IllegalArgumentException if first is null or empty
+     */
+    public static Path get(final String first, final String... more) {
+        return Paths.get(checkNotEmptyAfterTrimParam("first", first), more);
```
dmlloyd, line 55: Same comment as above.

io/src/main/java/io/smallrye/common/io/PathUtils.java
```diff
@@ -0,0 +1,2139 @@
...
+    /**
+     * A wrapper around {@link java.nio.file.Paths#get(URI)} with explicit null
+     * check.
+     *
+     * @param uri for the path, must be not null and a valid path
+     * @return the path constructed out of the uri
+     * @throws IllegalArgumentException if uri is null
+     */
+    public static Path get(final URI uri) {
+        return Paths.get(checkNotNullParam("uri", uri));
```
dmlloyd, line 67: I believe that `Paths.get()` also does a `null` check; why wrap it?

io/src/main/java/io/smallrye/common/io/PathUtils.java
```diff
@@ -0,0 +1,2139 @@
...
+    /**
+     * A wrapper around {@link java.nio.file.Files#newDirectoryStream(Path)} with
+     * explicit null check and unchecked {@code java.io.UncheckedIOException}.
+     *
+     * @param dir to the directory to list, must be not null.
+     * @return the elements in the directory as DirectoryStream
+     * @throws IllegalArgumentException if the path is {@code null}
+     * @throws UncheckedIOException around original IOExceptions
+     */
+    public static DirectoryStream<Path> newDirectoryStream(final Path dir) throws UncheckedIOException {
```
dmlloyd, line 79: I personally don't much like wrapping with `UncheckedIOException`.  We don't do that in just about any other place that I can think of. How would you justify this feature?

io/src/main/java/io/smallrye/common/io/PathUtils.java
```diff
@@ -0,0 +1,2139 @@
...
+            }
+            throw new UncheckedIOException(deleteFailed);
+        }
+        return path;
+    }
+
+    /**
+     * Recursive deletion of directories content, including subfolders.
+     *
+     * Adopted from Google Guava.
```
dmlloyd, line 751: Is there a license concern here?

io/src/main/java/io/smallrye/common/io/PathUtils.java
```diff
@@ -0,0 +1,2139 @@
...
+     *
+     * @param path to delete, must be not null, a path, and exists.
+     * @return the path to delete
+     * @throws UncheckedIOException around original IOExceptions and if not exists
+     *         path
+     * @see <a href=
```
dmlloyd, line 757: The content of `@see` is a member reference; it's not intended for external references.

io/src/main/java/io/smallrye/common/io/PathUtils.java
```diff
@@ -0,0 +1,2139 @@
...
+                deleteFailed.addSuppressed(e);
+            }
+            throw new UncheckedIOException(deleteFailed);
+        }
+        return path;
+    }
+
+    /**
+     * Recursive deletion of directories content, including subfolders.
+     *
```
dmlloyd, line 750: You should put a `<p>` when you want to separate doc paragraphs.

io/src/main/java/io/smallrye/common/io/PathUtils.java
```diff
@@ -0,0 +1,2139 @@
...
+     *         path
+     * @see <a href=
+     *      "https://github.com/google/guava/blob/master/guava/src/com/google/common/io/MoreFiles.java">com.google.common.io.MoreFiles</a>
+     */
+    public static Path deleteDirectoryContents(Path path) throws UncheckedIOException {
+        return doDeleteDirectoryContents(checkIsDirectoryNoFollowLinks(path));
+    }
+
+    static Path doDeleteDirectoryContents(Path path) throws UncheckedIOException {
+        Collection<IOException> exceptions = null; // created lazily if needed
```
dmlloyd, line 765: This is minor but you could have this variable be of type `FileSystemException` and pass that in to the nested methods rather than adding the suppressed exceptions to a collection only to iterate them all right away.

io/src/main/java/io/smallrye/common/io/FileUtils.java
```diff
@@ -0,0 +1,565 @@
...
+     * A wrapper around {@link java.nio.file.Paths#get(String, String...)} with
+     * explicit null check.
+     *
+     * @param first part of the path, must be not null, not empty and a valid path
+     * @param more optional parts of the path
+     * @return the path constructed out of first and more
+     * @throws IllegalArgumentException if first is null or empty
+     */
+    public static File get(final String first, final String... more) {
+        return Paths.get(checkNotEmptyAfterTrimParam("first", first), more).toFile();
```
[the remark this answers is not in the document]
boris-unckel: Thanks for your valuable time to review, I really appreciate that.

- Trimming whitespace at all: On both Windows and Linux/Unixiod systems path elements with leading or trailing space cause trouble. The method does not clean whitespace inside the names. The freedom to use leading/trailing whitespace is still there, but the default tries to avoid it. This is only for Strings the case, all other objects (Path, File, URI) allow it. I have documented the reason with links to detailed documentation in the class javadoc.
- Consistency between "first" and "more" parameter: thanks for the catch, I've fixed it.

io/src/main/java/io/smallrye/common/io/FileUtils.java
boris-unckel: Yes, this is intended. (No)FollowLinks is important to understand erroneous behaviour, this makes it directly visible in stackstraces and the developer will care, because even autocompletion shows directly the alternatives.
The copy options with it's multiple combinations would blow the code, I assume someone who is looking for "replace" instead of getting a "FileAlreadyExistsException" will start to look at the docs or google stackoverflow.
I have documented the approach in the class javadoc, including links.

io/src/main/java/io/smallrye/common/io/FileUtils.java
boris-unckel: Removed.

io/src/main/java/io/smallrye/common/io/PathUtils.java
boris-unckel: Same fix.

io/src/main/java/io/smallrye/common/io/PathUtils.java
boris-unckel: I have checked that and could not find a explicit null check. The intention is to have a Exception with the parameter name. I don't have a personal favour for NullPointer vs. IllegalArgument but I dislike the anonymous, fail late ones.

io/src/main/java/io/smallrye/common/io/PathUtils.java
boris-unckel: Fixed, moved to text.

io/src/main/java/io/smallrye/common/io/PathUtils.java
boris-unckel: Fixed.

io/src/main/java/io/smallrye/common/io/PathUtils.java
boris-unckel: No, Google Guava is licensed under the Apache v2 license. [https://github.com/google/guava/blob/master/COPYING](https://github.com/google/guava/blob/master/COPYING)

io/src/main/java/io/smallrye/common/io/PathUtils.java
boris-unckel: Before I started with any of the coding I have searched for places which might be worth to be refactored from classic java.io to java.nio. Many follow a pattern with private utility methods declaring a throws IOException (without Exception handling) and a major place where no Exception handling logic appears but to rethrow it as unchecked Exception.
You're right that currently the usage is different. I think the basic reason is backward compatibility of the JDK and missing 3rd party libraries supporting it.

I have added a section for this in the class javadoc.

io/src/main/java/io/smallrye/common/io/PathUtils.java
boris-unckel: The current approach allows lazy init. If I change that, the top-most exception would have a stacktrace top of a utility method or I can't lazy init.

io/src/main/java/io/smallrye/common/io/PathUtils.java
dmlloyd: I think the feature would *ideally* have to be justifiable within the context of SmallRye projects, or else be widely necessary by consumers of SmallRye projects, which is why I asked.  Unchecked versions of things is a specific enough need that I would imagine it to be framed specifically that way: e.g. `UncheckedPaths`, `UncheckedFiles` etc.  Also, specifically for methods like this one which return an object which is a subtype of `Closeable`, I'd expect a wrapper object which throws an unchecked exception on close e.g. `interface UncheckedDirectoryStream<Path> extends DirectoryStream<Path> { void close() throws UncheckedIOException; }`.

Put more simply, making unchecked versions of NIO methods is a separate concern from formalizing `null` checks according to a different scheme (though it would certainly be reasonable to perform `null` checks according to that scheme from such an API), and I'd expect the API to reflect that.

boris-unckel: Yes, something I didn't think of: It's not enough to provide "unchecked" utility methods, the underlying objects should behave in the same way.
This leads to a new problem: If UncheckedPath extends Path it's possible to provide different exception signatures, but a wrapper implementation can only throw either checked xor unchecked exceptions, not depending on a cast. All other objects (channels, buffer...) would still require checked ex behaviour.

In short: this is not ready to launch, neither in total nor in parts.

boris-unckel:
@dmlloyd Before I discard this PR I wanted to say thank you again for investing your valuable time. I really appreciate that. I have learned a lot,  about class design and an open source process.

PR #105, opened 2021-06-28
MNG-6964 is fixed upstream

dmlloyd:


dmlloyd:
Maybe don't merge until 3.8.2 is actually released so we can verify that the test works.

dmlloyd:
This needs to be reverted, sorry.

> Maybe don't merge until 3.8.2 is actually released so we can verify that the test works.

The test doesn't work.

PR #109, opened 2021-07-29
Build with Jakarta 9

radcortez:
Generates a build using Jakarta 9 dependencies by renaming the artifactId of the project and adding a `jakarta`name, replacing any `javax` to `jakarta` references in the source code and building new binaries from all the changes.

This allows creating a Jakarta 9 release without having to maintain two separate branches and keep the same version (only the artifatIds are different).

Example:

The SmallRye Common BOM GAV is `smallrye-common-bom` a new BOM for jakarta is generated as `smallrye-common-jakarta-bom`. The same rules apply to all other modules.


Emily-Jiang (approved):

xstefank (approved):

radcortez:
We are not going to use this approch. A separate branch will be added for the Jakarta namespace and updated as needed.

PR #129, opened 2022-02-02
Generate the Jandex index for the annotation module.

cescoffier:
It is required because it contains qualifiers and these qualifiers need to be in the index for Arc to find them.


Ladicek (approved):
Technically Quarkus could index the SmallRye Common Annotations JAR itself, but I can't think of any possible problem this PR could cause.

radcortez (approved):

PR #132, opened 2022-02-14
Provide utilities to properly use Vert.x duplicated context and Context locals

cescoffier:
CC @ozangunalp

### DIFF
--- a/vertx-context/pom.xml
+++ b/vertx-context/pom.xml
@@ -1,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.smallrye.common</groupId>
+        <artifactId>smallrye-common-parent</artifactId>
+        <version>1.9.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>smallrye-common-vertx-context</artifactId>
+    <name>SmallRye Common: Vert.x Context Utilities</name>
+    <description>A set of utility classes to ease the usage of Vert.x context locals and duplicated contexts.</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-core</artifactId>
+            <version>${version.vertx}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.smallrye.common</groupId>
+            <artifactId>smallrye-common-constraint</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-junit5</artifactId>
+            <version>${version.vertx}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.jboss.jandex</groupId>
+                <artifactId>jandex-maven-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+
+    <profiles>
+        <profile>
+            <id>coverage</id>
+            <properties>
+                <argLine>@{jacocoArgLine}</argLine>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.jacoco</groupId>
+                        <artifactId>jacoco-maven-plugin</artifactId>
+                        <executions>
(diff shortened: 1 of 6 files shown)


vertx-context/src/test/java/io/smallrye/common/vertx/ContextLocalsTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        vertx.runOnContext(x -> {
+            Assertions.assertThatThrownBy(() -> {
+                ContextLocals.remove("foo");
+            }).isInstanceOf(UnsupportedOperationException.class);
+
+            latch.countDown();
+        });
+
+        Assertions.assertThat(latch.await(100, TimeUnit.MILLISECONDS)).isTrue();
```
Ladicek, line 59: Nice way to propagate assertion status from the event loop thread to the main thread :-)

vertx-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Gets the value from the context local associated with the given key.
+     * If there is no associated value, it returns the given default.
+     *
+     * @param key the key, must not be {@code null}
+     * @param def the default value returned if there is no associated value with the given key.
+     *        Can be {@code null}
+     * @param <T> the expected type of the associated value
+     * @return an optional containing the associated value if any, the given default otherwise.
```
Ladicek, line 48:
```suggestion
     * @return the associated value if any, the given default otherwise.
```

Ladicek (approved):
Looks fine, just a minor correction of a javadoc.

vertx-context/src/test/java/io/smallrye/common/vertx/ContextLocalsTest.java
Ladicek: I'd be +1 to move to JUnit 5, obviously.

cescoffier:
Note: we are still collecting use cases from reactive messaging. So do not merge yet.

vertx-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
cescoffier: Thanks! Fixed!

vertx-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
```diff
@@ -0,0 +1,85 @@
+package io.smallrye.common.vertx;
```
ozangunalp, line 1: 👍  I think it is not necessary to add accessor for localContextData

vertx-context/src/main/java/io/smallrye/common/vertx/VertxContext.java
```diff
@@ -0,0 +1,88 @@
...
+     * If the method is called from a Vert.x thread, it retrieves the current context and checks if it's a duplicated
+     * context. Otherwise, it returns false.
+     *
+     * @return {@code true} if the method is called from a duplicated context, {@code false} otherwise.
+     */
+    public static boolean isOnDuplicatedContext() {
+        Context context = Vertx.currentContext();
+        return context != null && isDuplicatedContext(context);
+    }
+
```
ozangunalp, line 87: I think we would need a couple more methods :
- `@Nullable Context createDuplicatedContext()` : if on vertx context, create new duplicate context, otherwise null
- `@Nullable Context getOrCreateDuplicatedContext()` : if on vertx context, getOrCreateDuplicate, otherwise null
- `Context parentContext(Context context)` : return unwrap if duplicated context, otherwise itself

ozangunalp (changes requested):

vertx-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
cescoffier: Do you mean the method returning the map?

vertx-context/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
```
radcortez, line 37: Feel free to move this to JUnit 5.

vertx-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
ozangunalp: Yes, but I don't think that it is needed.

cescoffier: Agreed, I was thinking about it. 
We will see if we have a use case (it's easy to add, not to remove)

vertx-context/pom.xml
cescoffier: Done!

vertx-context/src/main/java/io/smallrye/common/vertx/VertxContext.java
cescoffier: Methods added (with slightly different names)

vertx-context/src/test/java/io/smallrye/common/vertx/ContextLocalsTest.java
cescoffier: I've switched to Vert.x Junit 5.

ozangunalp (approved):

radcortez (approved):

PR #140, opened 2022-03-18
Introduce reference support submodule

dmlloyd:
This would replace usages of `wildfly-common`'s `ref` subpackage.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,7 @@
         <module>io</module>
         <module>net</module>
         <module>os</module>
+        <module>ref</module>
         <module>version</module>
         <module>vertx-context</module>
     </modules>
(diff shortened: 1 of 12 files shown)


radcortez (dismissed):

dmlloyd:
Updated to use signed commits.

dmlloyd:
I don't understand the Windows failure. It's happening during BOM generation.

radcortez:
Strange, the previous build seems ok. I don't see a reason for that error either.

dmlloyd:
The `memento` method and its corresponding class were added in Maven 3.3.0, according to the JavaDoc.

dmlloyd:
I can only speculate that the Maven version on the Windows node was recently updated, and that `sundr` isn't compatible. I'm trying to verify that now.

dmlloyd:
Confirmed. Works with `3.8.2`, fails with `3.8.5`.

dmlloyd:
https://github.com/sundrio/sundrio/issues/309

radcortez (approved):

radcortez:
> Here's an idea for a workaround [using Maven wrappers](https://github.com/actions/setup-java/issues/40#issuecomment-1000724522).

Updated the workflows to use the maven wrapper, so the build passes now.

Feel free to merge the PR :)

PR #146, opened 2022-04-04
Add a module-info for smallrye-common-annotation

jponge:


jponge:
Ok, there are problems

PR #152, opened 2022-04-05
Ignore mvnw files

dmlloyd:


dmlloyd:
I don't know what's wrong with CI on this one. But anyway it just touches `.gitignore` so it should be fine...

radcortez (approved):

radcortez:
It is ignored in the CI :)

https://github.com/smallrye/smallrye-common/blob/main/.github/workflows/build.yml#L10

PR #153, opened 2022-04-05
Require Java 11 (minimum) at run time

dmlloyd:
Flatten all pre-11 source directories.


radcortez:
> Any thoughts on this?

Can it wait for SR Commons 2.0 (with the Jakarta move?)

dmlloyd:
> > Any thoughts on this?
> 
> Can it wait for SR Commons 2.0 (with the Jakarta move?)

Maybe; could you remind me what the timeline for that is expected to be?

radcortez:
We already have RCs for SR Commons 2.0. We can do the Final anytime, but we are delaying it until Wildfly / Quarkus also go Final with the Jakarta update. At this point, it seems that Jakarta 10 will go out by the end Q2, so maybe we need a Final in the next few weeks for Wildfly.

dmlloyd:
We should also make sure #158 is done before releasing; all we're waiting for now is a `jboss-logging` release.

radcortez (dismissed):

radcortez:
Let's hold this for SR Commons 2.0.

radcortez:
This PR is merged every time we update the `jakarta` branch.

radcortez:
Rebased in the `jakarta` branch and added to `main`.

PR #155, opened 2022-04-05
Add `module-info` to most (not all) submodules

dmlloyd:
This adds `module-info` to every submodule which falls under one of these categories:

* has no dependencies
* has dependencies on other artifacts from this project which have `module-info`
* has dependencies on artifacts with established module identifiers

The `module-info` descriptor is generated by my `module-info` Maven plugin. This is because the operation of that plugin is very simple and predictable (to me), and it produces a `module-info.class` file directly without trying to pass an intermediate `module-info.java` through `javac`, which doesn't work well. Another reason is that the generated `module-info.class` has a complete package list and version, which `javac` does not add (normally the `jar` tool handles this but Maven doesn't use it).

There are a few modules which depend on JBoss Logging or another dependency which does not yet have a `module-info` descriptor:

* `expression`
* `net`
* `version`
* `vertx-context`

One temporary solution for the first three could be to throw on a dependency for `org.jboss.logging` and let it break until we get it sorted out in that project with a release.


dmlloyd:
I'm looking into the Windows failure now.

radcortez (approved):

radcortez:
@jponge will this work for you?

jponge:
Let me check 😃

jponge:
Just tested locally, I can resolve `io.smallrye.common.annotation` fine 👍 

Thanks @dmlloyd @radcortez

dmlloyd:
See also https://github.com/jboss-logging/jboss-logging/pull/45

PR #193, opened 2022-10-21
Expression default value with DOUBLE_COLON

radcortez:


expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -299,13 +299,21 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
...
+                                    itr.next();
+                                    if (itr.hasNext() && itr.peekNext() == ':') {
+                                        // TP 7
+                                        // OK actually the whole thing is really going to be part of the key
+                                        // Best approach is, rewind and do it over again, but without end-on-colon
+                                        itr.rewind(start);
+                                        keyNode = parseString(itr, !flags.contains(Flag.NO_RECURSE_KEY), true, false, flags);
+                                        list.add(new ExpressionNode(general, keyNode, Node.NULL));
+                                    } else {
+                                        // TP 8
```
dmlloyd, line 312: This should be e.b. `TP 7.5` - or maybe `8.5` - the mnemonic means "test point" so that we can correlate tests with decision points in the NFA, even if they're similar. Then the corresponding test should be named accordingly so that it can be correlated.

expression/src/test/java/io/smallrye/common/expression/ExpressionTestCase.java
```diff
@@ -551,4 +551,22 @@ public void testPoint46() {
             assertEquals("", c.getExpandedDefault());
         }), "Should expand to empty string");
     }
+
+    @Test
+    void doubleColonFlagAndDefault() {
```
dmlloyd, line 556: Should be `testPoint7_5` or whatever to correlate above. Also missing are the following cases:

* `foo.bar::baz:something`
* `foo.bar::baz::something`
* `foo.bar:baz::something`
* `foo.bar:baz:something`

With, and without, `DOUBLE_COLON` set.

radcortez: I was wondering what should `foo.bar::baz:something` expand to. I think it makes sense that the resulting key is the entire expression since the `DOUBLE_COLON` flag states that the entire content is encoded into the key, but I can see people expecting to encode only `foo.bar::baz` in the key and use the default `something`.

dmlloyd (approved):

PR #190, opened 2022-10-17
Avoid triggering type cache pollution on VertxContext.isDuplicatedContext

Sanne:
This is a strong performance enhancement.
Using [@franz1981's new agent ](https://github.com/franz1981/type-pollution-agent) we can now identify suboptimal code in our libraries when doing `instanceof` (or comparable, such as casts) operations which trigger [JDK-8180450](https://bugs.openjdk.org/browse/JDK-8180450).

It would seem we have one of such cases in method `VertxContext.isDuplicatedContext` , the related portion of the report being:

```
--------------------------
2:	io.vertx.core.impl.DuplicatedContext
Count:	24323910
Types:
	io.vertx.core.impl.ContextInternal
	io.vertx.core.Context
Traces:
	io.smallrye.common.vertx.VertxContext.isDuplicatedContext(VertxContext.java:114)
		class: io.vertx.core.Context
		count: 11809285
	io.smallrye.common.vertx.VertxContext.isDuplicatedContext(VertxContext.java:115)
		class: io.vertx.core.impl.ContextInternal
		count: 10516203
	io.vertx.core.http.impl.Http1xServerConnection.handleMessage(Http1xServerConnection.java:135)
		class: io.vertx.core.impl.ContextInternal
		count: 509081
	io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:561)
		class: io.vertx.core.impl.ContextInternal
		count: 505264
	io.vertx.ext.web.impl.RoutingContextImpl.getEndHandlers(RoutingContextImpl.java:508)
		class: io.vertx.core.impl.ContextInternal
		count: 494211
	io.vertx.core.streams.impl.InboundBuffer.<init>(InboundBuffer.java:95)
		class: io.vertx.core.impl.ContextInternal
		count: 489866
--------------------------
```

As one can see, the common pattern is to cast an instance of `io.vertx.core.impl.ContextInternal` to  `io.vertx.core.impl.ContextInternal`; there's only one exception here perturbing the cache: this method is implicitly casting to its obvious interface `io.vertx.core.Context`.

The fix is simple: consistently cast to `io.vertx.core.impl.ContextInternal` exclusively.

PR #224, opened 2023-05-23
Additional type pollution fix

Sanne:
Similar to #190 , this looks rather trivial but has severe performance implications.

I wonder if we shouldn't delete all uses of `Assert.checkNotNullParam`, it seems rather dangerous.

### DIFF
--- a/vertx-context/src/main/java/io/smallrye/common/vertx/VertxContext.java
+++ b/vertx-context/src/main/java/io/smallrye/common/vertx/VertxContext.java
@@ -52,8 +52,10 @@ public class VertxContext {
      * @see #getOrCreateDuplicatedContext(Context)
      */
     public static Context getOrCreateDuplicatedContext(Vertx vertx) {
-        Vertx actual = Assert.checkNotNullParam("vertx", vertx);
-        Context context = actual.getOrCreateContext(); // Creates an event loop context if none
+        //Intentionally avoid to reassign the return from checkNotNullParam
+        //as it will trigger JDK-8180450
+        Assert.checkNotNullParam("vertx", vertx);
+        Context context = vertx.getOrCreateContext(); // Creates an event loop context if none
         if (isDuplicatedContext(context)) { // Also checks that the context is not null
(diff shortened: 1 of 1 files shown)


franz1981:
Avoiding the cast by returning void will work too

radcortez (approved):

Sanne:
> Avoiding the cast by returning void will work too

right that's essentially what I've done here: ignore the return type.  Perhaps that would be a good change to the `Assert.checkNotNullParam` so to avoid the problem in future.. WDYT @radcortez ?

radcortez:
> right that's essentially what I've done here: ignore the return type. Perhaps that would be a good change to the `Assert.checkNotNullParam` so to avoid the problem in future.. WDYT @radcortez ?

Sure. We probably need to deprecate that method and introduce a new one.

Ladicek:
Something like `this.foo = Objects.requireNonNull(foo)` (or `Assert.checkNotNullParam` for that matter) is a very common idiom in constructors. I have no idea why it was used here (and I'm obviously +1 for avoiding it), but removing the `Assert.checkNotNullParam()` method is not gonna remove the usages of this idiom. The `Objects.requireNonNull()` method exists in the JDK, and I'm pretty sure a lot of other projects have their own copy as well (I know I do in SmallRye Fault Tolerance, for example).

dmlloyd:
I have to also add a 🛑 to this line of reasoning... what we're working around here is a JVM bug which people are working on fixing, whereas an API change is much more long-term. As @Ladicek pointed out this is a commonly used pattern so changing the API isn't really warranted.

I'm OK with some simple/localized band-aid type fixes for the type pollution issue to stop the bleeding but I am against significant reengineering or API redesigning to work around it. Once the issue is fixed, I feel we will regret the amount of effort we spent on the problem which could have been spent in other areas, and we will also regret how we hobbled our own programming models for a short-term gain.

Sanne:
Could this get merged?

Sanne:
Thanks!

PR #238, opened 2023-08-25
Allow reaper threads to be started at run time

dmlloyd:
This does not contain the code to actually implement the build-time feature since (I think) we do not have a way to do this in this project. But this will be coupled with a change in Quarkus which registers the `io.smallrye.common.ref.References.ReaperThread` class to be reinitialized at run time, causing the reaper threads to start at run time. This avoids the "thread started on build time heap" error that is seen with the log manager update.


geoand (approved):

geoand (approved):
LGTM

ref/pom.xml
```diff
@@ -17,6 +17,10 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>smallrye-common-constraint</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.graalvm.sdk</groupId>
+            <artifactId>graal-sdk</artifactId>
+        </dependency>
```
Ladicek, line 23: Not an expert, but this might be concerning for some consumer projects like WildFly, maybe?

dmlloyd: The dependency has `provided` scope. If it is not present in the run time, the exception-handling path should take over and `false` will be returned for `isBuildTime`.

The `smallrye-common-net` subproject also uses this dependency in `provided` scope.

Ladicek: Ouch, I didn't look into the parent. Indeed `provided` scope makes sense.

Ladicek (approved):

PR #247, opened 2023-09-26
Create one Reference Reaper thread

gastaldi:
- This creates a single Reference Reaper thread instead of 3 (as discussed in
https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/Reference.20Reaper.201-3)

### DIFF
--- a/ref/src/main/java/io/smallrye/common/ref/References.java
+++ b/ref/src/main/java/io/smallrye/common/ref/References.java
@@ -4,7 +4,6 @@ import static java.security.AccessController.doPrivileged;
 
 import java.lang.ref.ReferenceQueue;
 import java.security.PrivilegedAction;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import org.graalvm.nativeimage.ImageInfo;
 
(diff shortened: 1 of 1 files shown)


dmlloyd (approved):

ref/src/main/java/io/smallrye/common/ref/References.java
```diff
@@ -32,11 +31,8 @@ static ReferenceQueue<Object> getReaperQueue() {
             if (isBuildTime()) {
                 // do nothing (class should be reinitialized)
             } else {
-                final AtomicInteger cnt = new AtomicInteger(1);
-                final PrivilegedAction<Void> action = () -> startThreadAction(cnt.getAndIncrement());
-                for (int i = 0; i < 3; i++) {
-                    doPrivileged(action);
-                }
+                final PrivilegedAction<Void> action = () -> startThreadAction(1);
```
franz1981, line 34: @gastaldi you can use an anonymous class, if the code style is not terribly affected, wdyt @dmlloyd @Sanne ?

gastaldi: Would that change anything performance-wise? I can change it if necessary

dmlloyd: I doubt it would make a difference; this code is only run once at startup. We still haven't shown that there is a measurable difference.

franz1981: Good points!

PR #268, opened 2023-12-06
Switch from `maven-artifact` to `maven-resolver` version scheme

### LINKED ISSUE #267: 3.6.0.CR1 considered newer than 3.6.0
Hey,

I tried to use `smallrye-common-version` for our release scripts and things went wrong because `3.6.0.CR1` is considered newer than `3.6.0`.

I switched back to using Maven Artifact as these two versions are sorted properly with `ComparableVersion`.

The algorithm in `MavenVersionScheme` is not exactly trivial so I will refrain from breaking things :).

/cc @dmlloyd

dmlloyd:
Fixes #267

### DIFF
--- a/version/src/test/java/io/smallrye/common/version/MavenVersionTest.java
+++ b/version/src/test/java/io/smallrye/common/version/MavenVersionTest.java
@@ -2,23 +2,24 @@ package io.smallrye.common.version;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-import org.apache.maven.artifact.versioning.ComparableVersion;
+import org.eclipse.aether.util.version.GenericVersionScheme;
+import org.eclipse.aether.version.InvalidVersionSpecificationException;
+import org.eclipse.aether.version.Version;
 import org.junit.jupiter.api.Test;
 
 @SuppressWarnings("SpellCheckingInspection")
 public class MavenVersionTest {
 
     @Test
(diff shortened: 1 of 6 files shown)


version/src/test/java/io/smallrye/common/version/MavenVersionTest.java
```diff
@@ -49,20 +50,26 @@ public void testMavenCompare() {
...
-        assertEquals(1, VersionScheme.MAVEN.compare("1-0.alpha", "1"));
+        checkMavenConsistency("1-0.alpha", "1");
         // https://issues.apache.org/jira/browse/MNG-6964
-        //checkMavenConsistency("1-0.beta", "1");
-        assertEquals(1, VersionScheme.MAVEN.compare("1-0.beta", "1"));
+        checkMavenConsistency("1-0.beta", "1");
+        // #267
+        checkMavenConsistency("3.6.0.CR1", "3.6.0");
+        checkMavenConsistency("3.6.0.FINAL", "3.6.0");
+        checkMavenConsistency("3.6.0.SP1", "3.6.0");
```
gsmet, line 59: Will `SP1` come before or after `3.6.0`?

dmlloyd: `CR1` comes before zero, `FINAL` is equal to zero, and `SP1` is greater than zero.

PR #278, opened 2024-01-18
Add CPU detection

dmlloyd:
As we start to use FFM, it may be useful to have a centralized and reliable means of detecting information about the current CPU architecture.


radcortez (approved):

dmlloyd:
Added a test and CI integration to verify that host CPU detection works.

PR #279, opened 2024-01-22
Provide API to construct IP addresses from host name and segments

dmlloyd:
Also, use the new API to construct the global IP address constants, to avoid unnecessary `String` construction.

Supercedes #276.


radcortez:
As a reference, here is the flamegraph for this PR:

Screenshot 2024-01-23 at 15 21 42

Which seems better than the current main:
https://github.com/smallrye/smallrye-common/pull/276#issuecomment-1904496104

radcortez (approved):

PR #293, opened 2024-03-25
Support multiple nesting of JAR files

### LINKED ISSUE #289: Doesn't work in Spring Boot jar
Spring BootJar has complex layout and special spring ClassLoader to make it work
SmallRye Config doesn't work in Spring Application packed as spring bootJar.

I have found, the problem is in your class
`io.smallrye.common.classloader.ClassPathUtils`

Full explanation and fix:
https://github.com/smallrye/smallrye-config/issues/1101

PS: more advanced research and fix are welcome 🙏

dmlloyd:
* Make sure that `jar:file:///...` are properly formed (the sub-URL must have a scheme of `file`)
* When there is no `!` in the JAR URL, treat the file as a JAR and process its root path
* Otherwise recursively resolve each segment which is `!`-terminated

Fixes #289

### DIFF
--- a/classloader/src/main/java/io/smallrye/common/classloader/ClassPathUtils.java
+++ b/classloader/src/main/java/io/smallrye/common/classloader/ClassPathUtils.java
@@ -8,12 +8,12 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.nio.file.FileSystem;
-import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.spi.FileSystemProvider;
 import java.util.Enumeration;
+import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -105,6 +105,19 @@ public class ClassPathUtils {
         });
     }
 
+    private static final FileSystemProvider JAR_PROVIDER;
+
+    static {
+        final ClassLoader ccl = Thread.currentThread().getContextClassLoader();
+        try {
+            Thread.currentThread().setContextClassLoader(ClassLoader.getSystemClassLoader());
+            JAR_PROVIDER = FileSystemProvider.installedProviders().stream().filter(p -> p.getScheme().equals("jar")).findFirst()
+                    .orElseThrow();
+        } finally {
+            Thread.currentThread().setContextClassLoader(ccl);
+        }
+    }
+
     /**
      * Attempts to represent a resource as a local file system path to be processed by a function.
      * If a resource appears to be an actual file or a directory, it is simply passed to the function as-is.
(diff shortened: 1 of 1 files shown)


dmlloyd:
One thing I am unsure of is whether it is correct to treat URLs of form `jar:file:///some/path/something.jar` as valid (current behavior) or invalid.

magicprinc:
TestApp still fails 😭
https://github.com/magicprinc/SmallRyeConfig-SpringBoot/tree/fix2

I have put your classes inside the app removed jar with original classes.
You can debug and see details

dmlloyd:
Do you know what URL was input into the method?

magicprinc:
`jar:file:/D:/TEMP/SmallRyeConfig-SpringBoot/build/libs/SmallRyeConfig-SpringBoot-1.0.jar!/BOOT-INF/classes!/application.properties`

magicprinc:
https://github.com/magicprinc/SmallRyeConfig-SpringBoot/tree/fixBundled

url: `jar:file:/D:/TEMP/SmallRyeConfig-SpringBoot/build/libs/SmallRyeConfig-SpringBoot-1.0.jar!/BOOT-INF/classes!/application.properties`

file: `file:/D:/TEMP/SmallRyeConfig-SpringBoot/build/libs/SmallRyeConfig-SpringBoot-1.0.jar!/BOOT-INF/classes!/application.properties`

jar: `D:\TEMP\SmallRyeConfig-SpringBoot\build\libs\SmallRyeConfig-SpringBoot-1.0.jar`

localPath: `/BOOT-INF/classes!/application.properties`

dmlloyd:
Can you determine whether the sub-entry `/BOOT-INF/classes` inside of `file:///D:/TEMP/SmallRyeConfig-SpringBoot/build/libs/SmallRyeConfig-SpringBoot-1.0.jar` is a JAR directory, or is it a nested JAR file? I am assuming that `!` means to treat that entry as a JAR file but Spring Boot might not agree with that, but this would be an easy way to check.

magicprinc:
image

magicprinc:
image

magicprinc:
image

dmlloyd:
One more fix to prevent wrong path lookups if the `!` is followed by a `/`.

magicprinc:
Works!

jarPath: `D:\TEMP\SmallRyeConfig-SpringBoot\build\libs\SmallRyeConfig-SpringBoot-1.0.jar`
path: `/BOOT-INF/classes!/application.properties`
→
localPath: `/BOOT-INF/classes`
path: `/BOOT-INF/classes!/application.properties`

magicprinc:
BTW, sorry for a noob question...
Why do you need such heavy machinery (FileSystems, path parsing)
As far as I know Spring uses only ClassLoader#getResources and url.openStream()

dmlloyd:
That's a more complex question than it seems to be ;)

There are several answers though:

* Sometimes there is no class loader for the target path
* Sometimes the class loader does not have the necessary capabilities
* Sometimes you are using an API that is based on `Path`
* Probably more cases I'm not thinking of right now

magicprinc:
When is it possible to have snapshot/RC version in maven repo? 😥🙏

dmlloyd:
I'm just doing some final testing with Quarkus to make sure nothing has broken.

PR #312, opened 2024-05-10
Add detection for WSL

### LINKED ISSUE #311: Add support for detecting WSL
We have this in Quarkus:

```
    /**
     * @return {@code true} if the application is running in a WSL (Windows Subsystem for Linux) environment
     */
    public static boolean isWSL() {
        var sysEnv = System.getenv();
        return sysEnv.containsKey("IS_WSL") || sysEnv.containsKey("WSL_DISTRO_NAME");
    }
```

and I think it would make more sense in SmallRye Common (probably OS?).

Now, I don't know if this implementation is good enough or if we should follow what @dmlloyd suggested here: https://github.com/quarkusio/quarkus/pull/40497#discussion_r1596901248

dmlloyd:
Fixes #311

### DIFF
--- a/os/src/main/java/io/smallrye/common/os/Linux.java
+++ b/os/src/main/java/io/smallrye/common/os/Linux.java
@@ -1,0 +1,57 @@
+package io.smallrye.common.os;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Utilities pertaining to the Linux operating system.
+ */
+public final class Linux {
+    private Linux() {
+    }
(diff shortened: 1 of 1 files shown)


gsmet:
Maybe it's a stupid question but are we sure Microsoft is not shipping a distribution specific to Azure that might contain `microsoft` in `/proc/version`?

dmlloyd:
Based on my research, I don't *think* so; this seems to be the most common idiom for detection of WSL and nobody has mentioned it not working on Azure. That said, there's nothing quite like testing :-)

PR #313, opened 2024-05-19
Add "@see RunOnVirtualThread" to Blocking and NonBlocking javadoc

sixcorners:
Think this is a good idea?


cescoffier:
Yes. 

But it needs to explain that methods annotated with @RunOnVirtualThread are considered @Blocking.

PR #316, opened 2024-06-06
Resource API

dmlloyd:
Proposed resource API. Provides a uniform API for reading resources from the class path or filesystem.

Non-goals are presently including:

- Support for writing

### DIFF
--- a/resource/src/main/java/io/smallrye/common/resource/PathResource.java
+++ b/resource/src/main/java/io/smallrye/common/resource/PathResource.java
@@ -1,0 +1,127 @@
+package io.smallrye.common.resource;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.ByteBuffer;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.charset.Charset;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.nio.file.attribute.FileTime;
+import java.time.Instant;
+
+import io.smallrye.common.constraint.Assert;
+
+/**
+ * A resource corresponding to a {@link Path}.
+ */
+public final class PathResource extends Resource {
+    /**
+     * Only allow cached mapping when the resource is very large.
+     */
+    private static final long MAP_THRESHOLD = 128 << 20; // 128MiB
+    private final Path path;
+    private URL url;
+    private MappedByteBuffer mapped;
+
+    /**
+     * Construct a new instance.
+     *
+     * @param pathName the relative path name (must not be {@code null})
+     * @param path the path (must not be {@code null})
+     */
+    public PathResource(final String pathName, final Path path) {
+        super(pathName);
+        this.path = Assert.checkNotNullParam("path", path);
+    }
+
+    /**
+     * {@return the path of this resource}
+     */
+    public Path path() {
+        return path;
+    }
+
+    public URL url() {
+        URL url = this.url;
+        if (url == null) {
+            try {
+                url = this.url = path.toUri().toURL();
+            } catch (MalformedURLException e) {
+                throw new IllegalStateException("Unexpected URL problem", e);
+            }
+        }
+        return url;
+    }
+
+    public DirectoryStream<Resource> openDirectoryStream() throws IOException {
+        return new MappedDirectoryStream<>(
+                Files.newDirectoryStream(path),
+                p -> new PathResource(pathName() + '/' + p.getFileName().toString(), p));
+    }
+
+    public boolean isDirectory() {
+        return Files.isDirectory(path);
+    }
+
+    public InputStream openStream() throws IOException {
+        return Files.newInputStream(path);
+    }
+
+    public ByteBuffer asBuffer() throws IOException {
+        MappedByteBuffer mapped = this.mapped;
+        if (mapped == null) {
+            long size = size();
+            if (size >= MAP_THRESHOLD) {
+                // map the (large) file into memory
+                synchronized (this) {
+                    mapped = this.mapped;
+                    if (mapped == null) {
+                        try (FileChannel fc = FileChannel.open(path(), StandardOpenOption.READ)) {
+                            if (fc.size() > Integer.MAX_VALUE) {
+                                throw new OutOfMemoryError("Resource is too large to load into a buffer");
+                            }
+                            mapped = this.mapped = fc.map(FileChannel.MapMode.READ_ONLY, 0, size);
+                        }
+                    }
+                }
+            }
+        }
+        if (mapped == null) {
+            // just read the bytes
+            return ByteBuffer.wrap(Files.readAllBytes(path));
+        } else {
+            return mapped.duplicate();
+        }
+    }
+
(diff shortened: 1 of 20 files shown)


geoand:
Very cool!

I think a follow up with a loader API would be most useful too

classloader/src/main/java/io/smallrye/common/classloader/resource/InMemoryResource.java
```diff
@@ -0,0 +1,64 @@
+package io.smallrye.common.classloader.resource;
```
geoand, line 1: Should `classloader` really be part of the package?

Ladicek: Agree. I even doubt this should be in the `classloader` module, there should be an extra module for this IMHO.

dmlloyd: Well, I tried to get away with it :)

dmlloyd:
I'm making some changes based on wanting a resource loader API as well (adding a canonical relative path to `Resource` with supporting helpers). I'm going to make this particular submodule require 17+, since it is new code (there is no compatibility risk).

radcortez:
Once we get the new Resource Loader API, should we deprecate `ClassPathUtils` or rewrite it to use the new APIs?

dmlloyd:
Good question. I'll see if rewriting them with the new APIs is feasible and I'll probably have a better answer then.

dmlloyd:
OK, here's where we stand with this.

We have `Resource`, which represents a resource (obviously). You can open a stream from it, get its URL, get its size, get its last-modified time, and get its relative path (which is always canonical by an efficient algorithm, and thus cannot "escape" via `..`). If it's a directory, and if the impl supports it, you can iterate the resources in the directory. You can also get a buffer containing the entire contents of the resource, and also get the resource contents as a single string.

There are implementations for JAR file entries, URLs, `Path`, and memory-backed resources. The memory-backed resource API is future-proof relative to FFM and `MemorySegment`. The `Path`-backed API supports `mmap` of large resources.

Then there's the `ResourceLoader` interface, which allows you to find a resource given a path relative to the root of the loader. There are resource loader implementations for JAR files (future proof against a possible future port of `wildfly-common`'s "archive" implementation, which supports `mmap` of JAR files), `Path`, and URLs.

It should be possible to reimplement (or replace our usages of) `ClassPathUtil` with this API. However, that module requires Java 11, so we'd have to also bump up the minimum runtime version of that module to 17 (which should be OK since I don't think WildFly is using this API at all). I think a logical next step would be to consider doing so, and then after that, to consider adding a base `ClassLoader` implementation which can easily make use of resources and resource loaders in some efficient manner.

classloader/pom.xml
```diff
@@ -31,6 +31,10 @@
             <artifactId>shrinkwrap-impl-base</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye.common</groupId>
+            <artifactId>smallrye-common-constraint</artifactId>
+        </dependency>
```
geoand, line 37: Is this change still needed?

resource/src/main/java/io/smallrye/common/resource/PathResource.java
```diff
@@ -0,0 +1,107 @@
...
+
+import com.sun.nio.file.ExtendedOpenOption;
+
+import io.smallrye.common.constraint.Assert;
+
+/**
+ * A resource corresponding to a {@link Path}.
+ */
+public final class PathResource extends Resource {
+    private static final long MAP_THRESHOLD = 256 << 10; // 256KiB
```
geoand, line 25: Any particular reason for choosing this threshold?

resource/src/main/java/io/smallrye/common/resource/ResourceUtils.java
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * An efficient relative path string canonicalizer which avoids copying and allocation to the maximum possible extent.
+     * The canonical path has no {@code .} or {@code ..} segments,
+     * does not contain consecutive {@code /},
+     * and does not begin or end with a {@code /}.
+     *
+     * @param path the path name (must not be {@code null})
+     * @return the canonical equivalent path (not {@code null})
+     */
+    static String canonicalizeRelativePath(String path) {
```
geoand, line 42: Impressive :)

geoand (approved):
Very nice! 

I added a couple of small quesrions

classloader/pom.xml
dmlloyd: No, good catch.

resource/src/main/java/io/smallrye/common/resource/PathResource.java
dmlloyd: No, it's 100% gut feeling. That said, I'm a bit nervous about mapping resources on an open-ended basis. I think I might revisit the API so that the lifetime of the buffer is bounded. This way on Java 21+ we can actively unload the buffer once the call is complete. I'll have to think about it a little bit.

geoand (approved):

resource/src/test/java/io/smallrye/common/resource/JarFileResourceLoaderTests.java
```diff
@@ -0,0 +1,154 @@
...
+                assertTrue(iterator.hasNext());
+                assertEquals("dir1/file-stored.txt", iterator.next().pathName());
+                assertTrue(iterator.hasNext());
+                assertEquals("dir1/file-deflated.txt", iterator.next().pathName());
+                assertFalse(iterator.hasNext());
+            }
+        }
+    }
+
+    private static JarFileResourceLoader makeJar(Entry... entries) throws IOException {
```
radcortez, line 57: We have `shrinkwrap` as a test dependency in other modules to help with this if you prefer. We can also use it here.

radcortez:
> It should be possible to reimplement (or replace our usages of) `ClassPathUtil` with this API. However, that module requires Java 11, so we'd have to also bump up the minimum runtime version of that module to 17 (which should be OK since I don't think WildFly is using this API at all).

We use it in SmallRye Config (and maybe other SmallRye projects are using it too). We had some discussions about moving SmallRye to Java 17 a few months ago, and we decided to wait a little longer. We probably need to revisit that discussion.

radcortez (approved):

resource/src/test/java/io/smallrye/common/resource/JarFileResourceLoaderTests.java
dmlloyd: I prefer to keep it simple in this case, since these tests may require detailed control over the archive format for specific reasons.

dmlloyd:
> > It should be possible to reimplement (or replace our usages of) `ClassPathUtil` with this API. However, that module requires Java 11, so we'd have to also bump up the minimum runtime version of that module to 17 (which should be OK since I don't think WildFly is using this API at all).
> 
> We use it in SmallRye Config (and maybe other SmallRye projects are using it too). We had some discussions about moving SmallRye to Java 17 a few months ago, and we decided to wait a little longer. We probably need to revisit that discussion.

I can wait on `ClassPathUtil`; it's definitely not urgent to change that API. Another option we have is to move Quarkus over to the new API, and deprecate `ClassPathUtil` and leave it until nobody is using it anymore, and then just drop it in favor of the new API. But we don't have to do anything right away.

PR #317, opened 2024-06-22
Save StringConcat bytecode generation for one-time operation

franz1981:
This has popup as a bytecode generation operation which can both save memory and CPU time at startup, see

image

It both save the lambda generation and, most importantly, a useless string concatenation.

### DIFF
--- a/ref/src/main/java/io/smallrye/common/ref/References.java
+++ b/ref/src/main/java/io/smallrye/common/ref/References.java
@@ -31,7 +31,16 @@ public final class References {
             if (isBuildTime()) {
                 // do nothing (class should be reinitialized)
             } else {
-                final PrivilegedAction<Void> action = () -> startThreadAction(1);
+                final PrivilegedAction<Void> action = new PrivilegedAction<>() {
+                    @Override
+                    public Void run() {
+                        final ReaperThread thr = new ReaperThread();
+                        thr.setName("Reference Reaper #1");
+                        thr.setDaemon(true);
+                        thr.start();
+                        return null;
+                    }
+                };
                 doPrivileged(action);
             }
         }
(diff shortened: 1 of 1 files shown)


radcortez (approved):

dmlloyd:
Why not just eliminate the string concatenation and keep the lambda? Keeping the lambda means one less class to load from disk.

franz1981:
@dmlloyd I have no strong opinion on this, so I can do it like that as well

dmlloyd:
:+1: it's probably better to keep the fix down to the thing being fixed when possible.

ref/src/main/java/io/smallrye/common/ref/References.java
```diff
@@ -31,7 +31,16 @@ static ReferenceQueue<Object> getReaperQueue() {
...
+                final PrivilegedAction<Void> action = new PrivilegedAction<>() {
+                    @Override
+                    public Void run() {
+                        final ReaperThread thr = new ReaperThread();
+                        thr.setName("Reference Reaper #1");
+                        thr.setDaemon(true);
+                        thr.start();
+                        return null;
+                    }
+                };
```
gastaldi, line 43:
```suggestion
                final PrivilegedAction<Void> action = () -> {
                    final ReaperThread thr = new ReaperThread();
                    thr.setName("Reference Reaper #1");
                    thr.setDaemon(true);
                    thr.start();
                    return null;
                };
```

franz1981: I've avoided this because of saving the runtime bytecode generation of lambdas

franz1981:
@radcortez similar to the others we know about ^^

geoand:
I actually independently came up with https://github.com/smallrye/smallrye-common/pull/388

radcortez:
Closed in favor of #388.

PR #330, opened 2024-07-17
Introduce `VersionRange`

gastaldi:
This introduces a predicate that can be used for comparing if given a version is contained inside a version range pattern


gastaldi:
@dmlloyd @aloubyansky I introduced an utility class to check the range introduced in https://github.com/quarkusio/quarkus/pull/41908

version/src/main/java/io/smallrye/common/version/VersionRange.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A {@link VersionRange} is a predicate that tests if a version string is within a specified range.
+ */
+public class VersionRange implements Predicate<String> {
+
+    /**
+     * Range in format "[1.0,2.0)"
```
aloubyansky, line 15: We could probably borrow a more comprehensive version range impl from, for example, `org.apache.maven.artifact.versioning.VersionRange`?

gastaldi: Ah that's a good idea, I am not checking for overlaps or caching the ranges, for example

version/src/main/java/io/smallrye/common/version/VersionRange.java
```diff
@@ -0,0 +1,69 @@
...
+     */
+    private static final Pattern RANGE_PATTERN = Pattern.compile("([\\[\\(])(.*),(.*)([\\]\\)])");
+
+    private final VersionScheme versionScheme;
+    private final Bound lowerBound;
+    private final Bound upperBound;
+
+    public VersionRange(VersionScheme versionScheme, String rangePattern) {
```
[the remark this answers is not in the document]
dmlloyd: This might be nicer if it were a default method on `VersionScheme`. And, you could even make it use composable sub-predicates for arbitrary ranges, like this:

```java
public interface VersionScheme {
    // ...
    default boolean lt(String base, String other) {
        return compare(base, other) < 0;
    }
    // and le, gt, ge

    default Predicate<String> whenLt(String base) {
        return other -> lt(base, other);
    }
    // and whenLe, whenGt, whenGe
```

Then you can easily compose:

```java
    VersionScheme vs = VersionScheme.MAVEN;
    Predicate<String> isVersion1 = vs.whenGe("1").and(vs.whenLt("2"));
```

And then it's pretty trivial to convert the regex pattern result directly into predicates using this API.

version/src/main/java/io/smallrye/common/version/VersionRange.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Map;
+import java.util.WeakHashMap;
+import java.util.function.Predicate;
+
+/**
+ * A {@link VersionRange} is a predicate that tests if a version string is within a specified range.
+ */
+public class VersionRange implements Predicate<String> {
+
+    private static final Map<String, VersionRange> CACHE_SPEC = Collections.synchronizedMap(new WeakHashMap<>());
```
dmlloyd, line 15: I wouldn't bother with a cache, to be honest.

gastaldi: Gotcha, I reverted in a separate commit now

version/src/main/java/io/smallrye/common/version/VersionRange.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * A {@link VersionRange} is a predicate that tests if a version string is within a specified range.
+ */
+public class VersionRange implements Predicate<String> {
+
+    /**
+     * Range in format "[1.0,2.0)"
+     */
+    private static final Pattern RANGE_PATTERN = Pattern.compile("([\\[\\(])(.*),(.*)([\\]\\)])");
```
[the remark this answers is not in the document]
gastaldi: Thanks, but I removed the pattern completely

aloubyansky:
Looks nice @gastaldi, thanks! Could we also add tests involving version qualifiers? Right now the tests are limited to "Final" versions. So, for example, if a range starts with `[1.0,...]` then `1.0.0.Alpha3` shouldn't be accepted.

aloubyansky:
But `[1.0.a,..,]` should include `1.0.0.Alpha3`

aloubyansky:
We may want to test `SP` and `redhat` in the mix as well, it's a known troublesome combination for dependabot, for example.

aloubyansky:
The `SP` and `redhat` case would be more than what we currently need from this though. So shouldn't be a blocker if it appears to be problematic.

gastaldi:
@aloubyansky I added more tests, see if they make sense

version/src/test/java/io/smallrye/common/version/VersionRangeTest.java
```diff
@@ -0,0 +1,98 @@
...
+        assertFalse(versionRange.test("1.2.0.Alpha1"));
+    }
+
+    @Test
+    public void testQualifiers() {
+        VersionRange versionRange = VersionRange.createFromVersionSpec(VersionScheme.MAVEN, "[3.8,3.8.5)");
+        assertTrue(versionRange.test("3.8.4.SP1-redhat-00001"));
+        assertTrue(versionRange.test("3.8.4.SP2-redhat-00001"));
+        assertTrue(versionRange.test("3.8.4.redhat-00002"));
+        assertFalse(versionRange.test("3.8.5.redhat-00003"));
```
aloubyansky, line 95: Thanks, this looks good. What if you test `3.8.0.SP1-redhat-00001` for `[3.8.0.redhat-00001,)`?

gastaldi: Hm, that fails because the algorithm handles `SP1` as higher than `redhat` (alphabetical order). 

This returns `1`: 

```java
VersionScheme.MAVEN.compare("3.8.0.redhat-00001", "3.8.0.SP1-redhat-00001")
```

gastaldi: I wonder if we shouldn't have a specific Red Hat `VersionScheme` for cases like these

aloubyansky: It's ok, it's not a blocker, I was just wondering. Dependabot has the same issue. We can fix it later. It'd be good to fix it though :)

aloubyansky (approved):

version/src/test/java/io/smallrye/common/version/VersionRangeTest.java
aloubyansky: Maven resolver actually does handle it properly.

gastaldi: Hm, that sounds like a bug in `MavenVersionScheme`. WDYT @dmlloyd ?

version/src/main/java/io/smallrye/common/version/VersionRange.java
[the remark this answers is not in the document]
gastaldi: I'd leave that as an exercise for the reader 😉

version/src/test/java/io/smallrye/common/version/VersionRangeTest.java
dmlloyd: I believe that custom qualifiers come after predefined qualifiers (like `SP`) according to the Maven rules.

```
 $ jbang --main=org.eclipse.aether.util.version.GenericVersionScheme org.apache.maven.resolver:maven-resolver-util:1.9.6 1.redhat 1.sp
Display parameters as parsed by Maven Resolver (in canonical form and as a list of tokens) and comparison result:
1. 1.redhat -> 1.redhat; tokens: [1, redhat]
   1.redhat > 1.sp
2. 1.sp -> 1.sp; tokens: [1, 1]
```
```
$ jbang --main=org.eclipse.aether.util.version.GenericVersionScheme org.apache.maven.resolver:maven-resolver-util:1.9.6 3.8.0.redhat-00001 3.8.0.SP1-redhat-00001
Display parameters as parsed by Maven Resolver (in canonical form and as a list of tokens) and comparison result:
1. 3.8.0.redhat-00001 -> 3.8.0.redhat-00001; tokens: [3, 8, redhat, 1]
   3.8.0.redhat-00001 > 3.8.0.SP1-redhat-00001
2. 3.8.0.SP1-redhat-00001 -> 3.8.0.SP1-redhat-00001; tokens: [3, 8, 1, 1, redhat, 1]
```

dmlloyd: Make sure you aren't using the utilities in `maven-artifact` to test, because that is generally considered to be deprecated at this point (and IIUC is not used by Maven anymore).

version/src/main/java/io/smallrye/common/version/VersionRange.java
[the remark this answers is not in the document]
dmlloyd: I prefer it to be solved this way, please.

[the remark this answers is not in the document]
gastaldi: Introduced in [`950941d` (#330)](https://github.com/smallrye/smallrye-common/pull/330/commits/950941db041b72b1a0e97f95f972b7d36a52b62e)

dmlloyd:
There are two main problems remaining that I can see.

The first problem is that of the parser. It operates by searching the same string ranges many times (using `indexOf`), and comparing indices to see which things come first (if they are found). It is hard to prove that such a parser is in fact correct. Normally a parser should proceed from left to right (or, rarely, from right to left) and visit each character once, making a decision at that point. In this way it is easier to verify that in some given parsing state, the successor states are all valid and make sense.

The second problem is the restriction classes, which implement a parallel variety of predicate compared to the `whenXxx` methods that I had you add. The point of adding these methods would be to have only one kind of predicate. For example, if I give a version range of `(,1.2)` then I'd want that to be exactly identical to calling `whenLt("1.2")`. As a side effect of this duplication, we have a couple extra classes, one of which is visible to the user.

I'd made an example commit [here](https://github.com/dmlloyd/smallrye-common/commit/range) which you can feel free to use as you wish. This commit institutes a left-to-right parser and ensures there is only one way for the API to represent a range predicate.

One thing that this commit does _not_ do, which we can look into if you think this would be a valid use case, is validate that a given version range is indeed valid (other than a couple very opportunistic checks). The reason for this again is that the predicate methods should behave identically to the range-parsed predicates. That is, if the parser rejects `(2.0,1.0)` then the predicate methods should also reject `whenLt("2.0").and(whenGt("1.0"))`. In order for this to happen, the predicate methods would have to produce instances with a more comprehensive API (somewhat similar to your existing `VersionRestriction` class). I've left that off as something we could look into later, possibly. In my view, I think it would be totally valid for `(2.0,1.0)` to just yield `() -> false` (or something that behaves in an identical way), which is already what would happen if you combined the predicate methods (as they are currently defined) in an invalid way.

I hope this makes sense. Feel free to use some, all, or none of my example commit to get this over the line. Thanks!

gastaldi:
@dmlloyd loved it! I've pushed your commit to my branch for your appreciation

dmlloyd (approved):

PR #339, opened 2024-08-29
Cache current process handle

dmlloyd:
Fixes performance problem when creating log records or other things which frequently request the PID or process name.

See quarkusio/quarkus#42858 for more info.

### DIFF
--- a/os/src/main/java/io/smallrye/common/os/Process.java
+++ b/os/src/main/java/io/smallrye/common/os/Process.java
@@ -29,6 +29,10 @@ import java.util.List;
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 public final class Process {
+    private static final ProcessHandle current = doPrivileged((PrivilegedAction<ProcessHandle>) ProcessHandle::current);
+    private static final ProcessHandle.Info currentInfo = current.info();
+    private static final String name = doPrivileged((PrivilegedAction<String>) Process::computeProcessName);
+
     private Process() {
     }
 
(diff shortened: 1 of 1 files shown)


os/src/main/java/io/smallrye/common/os/Process.java
```diff
@@ -29,6 +29,10 @@
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 public final class Process {
+    private static final ProcessHandle current = doPrivileged((PrivilegedAction<ProcessHandle>) ProcessHandle::current);
```
gsmet, line 32: Do we actually need to keep this one around.
From what I can see, it's only used to get the PID?

I wonder if we should have a private `record` with the state (pid, process name, command) and only cache this?

`ProcessHandle` is out of our reach and I'm very nervous making it static.

dmlloyd: When you call `ProcessHandle.current()`, it's returning a static final field already. The object itself is a holder for a PID value (`long`)/start time (another `long`) combination, so generally it's safe to retain (especially _our own_ PID). 

What we're caching here is basically the side-effects of calling through `doPrivileged`, which is still required for WildFly for a little while longer. Once that's no longer needed by any downstream projects, we can replace it with calls to `ProcessHandle.current()`.

gsmet: OK, all good then, thanks for the details.

I have probably spent too much time looking at heap dumps and CL leaks, lately :).

radcortez (approved):

PR #344, opened 2024-09-13
Expression NO_$$ mode

radcortez:
Proposal of a new escape expressions mode (`NO_$$`) to help with:

- https://github.com/smallrye/smallrye-config/issues/1219
- https://github.com/smallrye/smallrye-config/issues/746

Activating `NO_$$` will effectively disable the `$$` escape used for expressions.

This also allows the removal of the workaround for MP Config:
https://github.com/smallrye/smallrye-config/blob/d447b0c346e39a09c6d07c7d5ce8186bd4154037/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java#L102-L123

### DIFF
--- a/expression/src/main/java/io/smallrye/common/expression/Expression.java
+++ b/expression/src/main/java/io/smallrye/common/expression/Expression.java
@@ -410,6 +410,9 @@ public final class Expression {
                             if (flags.contains(Flag.MINI_EXPRS)) {
                                 // TP 13
                                 list.add(new ExpressionNode(false, LiteralNode.DOLLAR, Node.NULL));
+                            } else if (flags.contains(Flag.NO_$$)) {
+                                list.add(LiteralNode.DOLLAR);
+                                itr.prev();
                             } else {
                                 // just resolve $$ to $
                                 // TP 14
@@ -493,10 +496,12 @@ public final class Expression {
                                         Node.NULL));
                                 start = itr.getNextIdx();
                                 continue;
-                            } else if (flags.contains(Flag.LENIENT_SYNTAX)) {
+                            } else if (flags.contains(Flag.LENIENT_SYNTAX) || flags.contains(Flag.NO_$$)) {
                                 // TP 26
                                 // just treat it as literal
-                                start = itr.getPrevIdx() - 1; // we can use 1 here because unicode '$' is one char in size
+                                list.add(LiteralNode.DOLLAR);
+                                itr.prev();
+                                start = itr.getNextIdx();
                                 continue;
                             } else {
                                 // TP 27
@@ -592,6 +597,10 @@ public final class Expression {
                                 case 'f': {
                                     // TP 39
                                     node = LiteralNode.FORM_FEED;
+                                    break;
+                                }
+                                case '$': {
+                                    node = LiteralNode.DOLLAR;
                                     break;
                                 }
                                 case '\\': {
@@ -694,6 +703,11 @@ public final class Expression {
          * character.
          */
         ESCAPES,
+        /**
+         * Escaping <code>$</code> with <code>$$</code> or <code>/$</code> only applies when <code>{</code> follows
+         * the initial escaped <code>$</code>.
+         */
+        NO_$$,
         /**
          * Treat expressions containing a double-colon delimiter as special, encoding the entire content into the key.
          */
--- a/expression/src/test/java/io/smallrye/common/expression/ExpressionTestCase.java
+++ b/expression/src/test/java/io/smallrye/common/expression/ExpressionTestCase.java
@@ -674,8 +674,8 @@ public class ExpressionTestCase {
         }), "SRCOM01000: Invalid expression syntax at position 0");
         assertThrows(IllegalArgumentException.class, () -> Expression.compile("a\\", ESCAPES).evaluate((c, b) -> {
         }), "SRCOM01000: Invalid expression syntax at position 0");
-        assertThrows(IllegalArgumentException.class, () -> Expression.compile("\\${foo}", ESCAPES).evaluate((c, b) -> {
-        }), "SRCOM01000: Invalid expression syntax at position 0");
+        assertEquals("${foo}", Expression.compile("\\${foo}", ESCAPES).evaluate((c, b) -> {
+        }));
 
         // NO_SMART_BRACES
         assertEquals("", Expression.compile("${foo}", NO_SMART_BRACES).evaluate((c, b) -> {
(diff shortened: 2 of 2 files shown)


radcortez:
One of the reasons was that it doesn't play well with the comma escape for multiple elements in arrays, like `array=cat,dog,${mouse},sea\\,turtle`

dmlloyd:
Sure, I can see that. However it is not possible to unambiguously discriminate in this case, so I think we may just have to "eat it" and require double-escaping for lists in order to truly fix the problem, or else come up with a single expression escaping scheme that also covers the list case (which is much harder than it would seem, since lists of lists are a thing).

dmlloyd:
Basically, this is a state machine (a DFA), and each state is well-defined (basically meaning, every character path is evaluated and tested for each state). By adding a boolean flag which sometimes changes the code path, the number of states has doubled. Unless every new state is tested (or proven not to be enterable) then there is a lacking of test coverage.

dmlloyd:
I think you're right about `$$` - that should be refactored to be a part of `MINI_EXPRS`, and having it be separate along with the lack of `ESCAPES` really causes confusion and problems. It'll break some things but I think the end result will be more consistent and easier to explain.

dmlloyd:
Of course then I'll have to find all the places where I said that `$$` is the official correct way to escape `$` and update that. 😢

radcortez:
> Basically, this is a state machine (a DFA), and each state is well-defined (basically meaning, every character path is evaluated and tested for each state). By adding a boolean flag which sometimes changes the code path, the number of states has doubled. Unless every new state is tested (or proven not to be enterable) then there is a lacking of test coverage.

I added the flag make the checks easiers, but I think it could be writtent without it.

> I think you're right about `$$` - that should be refactored to be a part of `MINI_EXPRS`, and having it be separate along with the lack of `ESCAPES` really causes confusion and problems. It'll break some things but I think the end result will be more consistent and easier to explain.

Ok, so we have two options: 

Either we go with the new expression mode and cover our current expected cases

or 

We turn on `ESCAPES` and make `$$` escape only for `MINI_EXPRS`. We know it breaks escaping commas for lists, which probably has a shallow usage (if any). We need to check if there are other cases in which it breaks.

radcortez:
> Of course then I'll have to find all the places where I said that `$$` is the official correct way to escape `$` and update that. 😢

Or we could still use it if `$` is immediately followed by a `{` with an ending `}`, and ignore it in all the other cases.

radcortez:
I've made a small change to remove the flag.

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -258,6 +262,25 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
                         // treat as plain content
                         continue;
                     }
+                    if (flags.contains(Flag.STRICT_ESCAPES)) {
+                        if (!itr.hasNext()) {
```
dmlloyd, line 266: This should all be done within the next state. Just as we don't want to add another state flag, we also don't want to introduce backtracking for a similar reason: the state machine should only have one way to reason about it if possible.

This logic should all fit into the `case '$'` substate below.

radcortez: Yes, I wanted to minimize changes to the current code, but I'll do that.

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -174,6 +174,15 @@ public static Expression compile(String string, Flag... flags) {
...
         Assert.checkNotNullParam("string", string);
         Assert.checkNotNullParam("flags", flags);
+
+        if (flags.contains(Flag.STRICT_ESCAPES) && flags.contains(Flag.ESCAPES)) {
+            throw Messages.msg.mutuallyExclusiveEscapeFlags();
+        }
+
+        if (flags.contains(Flag.STRICT_ESCAPES) && flags.contains(Flag.MINI_EXPRS)) {
+            throw Messages.msg.incompatibleMiniExpsStrictEscapes();
+        }
```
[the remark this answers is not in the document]
radcortez: Actually, they can coexist. I added tests to showcase it.

In my opinion, I thought it didn't make much sense to combine the modes, so I added the restrictions. I've now removed the restrictions.

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -407,18 +410,47 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
...
+                                            itr.prev();
+                                            break;
+                                        default:
+                                            list.add(new LiteralNode(itr.getStr(), start, itr.getNextIdx()));
+                                            break;
+                                    }
+                                } else {
+                                    list.add(new LiteralNode(itr.getStr(), start, itr.getNextIdx()));
+                                }
                             } else {
```
dmlloyd, line 430: Now that the semantics of the flag are being more clear, couldn't we instead add one extra `if`/`else` to the next part (between TP 13 and TP 14) where if `NO_$$` is set we enqueue a `$` and re-enter the "got `$`" state, and then just forget everything else?

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -485,15 +507,30 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
...
+                        case '\\': {
+                            if (flags.contains(Flag.NO_$$)) {
+                                if (itr.hasNext() && itr.peekNext() == '$') {
+                                    list.add(new LiteralNode(itr.getStr(), start, idx));
+                                    start = itr.getNextIdx();
+                                    itr.next();
+                                    continue;
+                                }
+                            }
+                        }
```
dmlloyd, line 519: This shouldn't be needed anymore AFAICT.

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -269,16 +272,16 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
                         start = itr.getNextIdx();
                         continue;
                     }
-                    // enqueue what we have acquired so far
-                    if (idx > start) {
-                        // TP 4
-                        list.add(new LiteralNode(itr.getStr(), start, idx));
-                    }
```
dmlloyd, line 276: This can be undone since the same piece should be in every possible subcase now.

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -406,10 +409,19 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
...
+                            }
                             // $$
                             if (flags.contains(Flag.MINI_EXPRS)) {
                                 // TP 13
                                 list.add(new ExpressionNode(false, LiteralNode.DOLLAR, Node.NULL));
+                            } else if (flags.contains(Flag.NO_$$)) {
+                                list.add(LiteralNode.DOLLAR);
+                                if (!itr.hasNext() || itr.peekNext() != '{') {
+                                    itr.prev();
+                                }
```
dmlloyd, line 424: We shouldn't need to care whether or not the next character is a `{` for this case. We should be acting in exactly the same way as the default branch here if `MINI_EXPRS` is off and `NO_$$` is present.

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -485,15 +507,30 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
...
+                            }
                             // $ followed by anything else
                             if (flags.contains(Flag.MINI_EXPRS)) {
                                 // TP 25
                                 list.add(new ExpressionNode(false, new LiteralNode(itr.getStr(), idx, itr.getNextIdx()),
                                         Node.NULL));
                                 start = itr.getNextIdx();
                                 continue;
-                            } else if (flags.contains(Flag.LENIENT_SYNTAX)) {
+                            } else if (flags.contains(Flag.LENIENT_SYNTAX) || flags.contains(Flag.NO_$$)) {
```
dmlloyd, line 533: Why do we care about `NO_$$` here?

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -551,7 +588,22 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
...
+                                list.add(new LiteralNode(itr.getStr(), start, escape - 1));
+                                list.add(new LiteralNode(itr.getStr(), escape, itr.getNextIdx()));
+                                start = itr.getNextIdx();
+                            } else if (itr.hasNext() && itr.peekNext() == '$') {
+                                list.add(new LiteralNode(itr.getStr(), start, escape));
+                                start = itr.getPrevIdx();
+                                itr.prev();
+                            }
+                        }
+                        continue;
```
dmlloyd, line 605: Why do we care about `NO_$$` here?

expression/src/main/java/io/smallrye/common/expression/Expression.java
radcortez: My idea was to still keep `$$` as an escape sequence in the presence of an expression, as:

`$${foo}` -> `${foo}`
`$$foo` -> `$$foo`

The goal is to minimize breaking user configuration

expression/src/main/java/io/smallrye/common/expression/Expression.java
radcortez: In the case of expressions like `$$foo`, when we reevalute the second '$' we fall in the `// $ followed by anything else` case.

expression/src/main/java/io/smallrye/common/expression/Expression.java
radcortez: The goal here was to support the MP escape sequence of `\$`, without turning on `ESCAPES`, which breaks commas separated values escapes.

expression/src/main/java/io/smallrye/common/expression/Expression.java
radcortez: The goal here was to support the MP escape sequence of \$, without turning on ESCAPES, which breaks commas separated values escapes.

expression/src/main/java/io/smallrye/common/expression/Expression.java
dmlloyd: I'm not sure that goal is as important as having a consistent parsing strategy. Whenever you make an exception case to escaping, then something breaks. You can either have it be something common but easy to fix, or rarer but harder to fix. If your escaping scheme is not general, then you have to hunt harder for the case that is not easy to fix but it's going to be there.

I think it's best to make the scheme as simple as possible, and just accept the breakage. We could potentially come up with a config migration scheme if needed, though I doubt it would be needed.

radcortez: Ok, so if we drop `$$` as an escape for this particular case, do we use `\$`?

dmlloyd: Yes, the new policy would be "`\` always escapes the character that comes after it; if you need to pass a `\` to the config layer then use `\\`". Then the only effect of `NO_$$` is that `$$` is not treated as a special sequence in any way unless `MINI_EXPRS` is turned on.

radcortez: Adding `ESCAPES`? Or we always add the `\$` escape to `NO_$$`?

By moving to `NO_$$`we are adding a braking change for configuration that relies on `$$` as the escape, which needs to change to `\$`. If we move to `ESCAPES`, this will also break comma escapes in comma-separated values.

I would prefer to keep the `ESCAPES` off and just add `\$` to `NO_$$`.

dmlloyd: If we add `\$` to `NO_$$` then we have a situation where `\` does different things based on context, and in order to get a `\` you'd need to do `\\` or something similar, and now you basically have an inconsistent version of `ESCAPES` because if you need `\` you say `\`, but if you need `\\` you have to say `\\\\`.

If, in config, we use `ESCAPES` and the simplified `NO_$$`, now we're 100% consistent: `\\` means `\`, `\$` means `$`, `\` followed by *anything* always means *something*, and `$` is an expression iff it is followed by `{`.

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -260,6 +260,9 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
                     }
                     // check to see if it's a dangling $
                     if (!itr.hasNext()) {
+                        if (flags.contains(Flag.NO_$$)) {
```
dmlloyd, line 263: What is the expected behavior here? IIRC this state is entered whenever we hit a `$` even if we've hit other things. So, I think we do not need to do anything special here because if `LENIENT_SYNTAX` is on, then we pass it through below and add the literal text from the start point to the end of the string. Otherwise it's an error. This is regardless of `NO_$$` AFAICT.

expression/src/main/java/io/smallrye/common/expression/Expression.java
```diff
@@ -485,6 +491,14 @@ private static Node parseString(Itr itr, final boolean allowExpr, final boolean
...
                             //throw Assert.unreachableCode();
                         }
+                        case '\\': {
+                            if (flags.contains(Flag.ESCAPES)) {
+                                list.add(LiteralNode.DOLLAR);
+                                itr.prev();
+                                start = itr.getNextIdx();
+                                continue;
+                            }
+                        }
```
dmlloyd, line 501: Why is this case needed? Isn't this handled by the `default` below?

expression/src/main/java/io/smallrye/common/expression/Expression.java
dmlloyd: OK, I guess this was a perf optimization. It might be cleaner if for this case we enqueue what we have so far (before the `$`), add a literal `$` item, and then continue from initial, which would flatten a couple of other cases too. (This might be similar to what you had before; if so, sorry about that... 🙁)

expression/src/main/java/io/smallrye/common/expression/Expression.java
radcortez: Leftovers from the previous implementation.

expression/src/main/java/io/smallrye/common/expression/Expression.java
radcortez: Not without your suggested change here: https://github.com/smallrye/smallrye-common/pull/344#discussion_r1783112429 :)

I didn't want to change it based on some of our previous conversations, but we can remove it if we change the other case.

expression/src/main/java/io/smallrye/common/expression/Expression.java
radcortez: No worries, all good :)

dmlloyd (approved):
Looks good, nice and minimal now. Thanks for your patience!

radcortez:
All good! Thank you for your patience! :)

PR #349, opened 2024-09-27
Move baseline to Java 17

### LINKED ISSUE #348: Move to Java 17 as a baseline requirement
It's time. (Is it time?)

dmlloyd:
Fixes #348, and some other minor issues as well

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,6 @@ jobs:
         with:
           distribution: temurin
           java-version: |
-            11
             17
             21
           cache: 'maven'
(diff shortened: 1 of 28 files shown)


radcortez (approved):

radcortez:
Please wait until next week for our SR meeting so we can also discuss this there.

dmlloyd:
Any update?

radcortez:
We can move once we make final releases to support MP 7.0. Not all implementations are required, but it probably wouldn't hurt to have all components to make a final release with 11 before moving to 17.

radcortez:
Maybe we should do a Parent update to 17 first, release, and then here?

dmlloyd:
The parent doesn't really need to change until there are no consumers left which rely on 11 AFAICT. I don't think it makes sense to block on it at any rate.

dmlloyd:
Now that 2.8.0 was released, what do we think about moving this forward?

dmlloyd:
Per conversations, I'm going ahead with this. If there is a specific problem, please let me know (I'm on Zulip).

PR #361, opened 2024-10-21
Add resource method to list code signers

dmlloyd:
This allows for secure (spec-compliant) class loader implementation by allowing correct protection domains to be defined.


dmlloyd:
Fixed a typo.

PR #368, opened 2024-10-30
Always use a URL stream handler for path-based resources

dmlloyd:
This avoids a potential problem where e.g. filesystem locks are taken by the JDK's caching implementation.

See quarkusio/quarkus#43940 for an example.


geoand (approved):

dmlloyd:
I think I want to add some tests for this, but I need to think about it a little first.

PR #375, opened 2024-11-19
Fix release build

dmlloyd:
* Fully modularize modules which did not previously have `module-info.yml`
* Disable JavaDoc legacy mode

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,6 @@
     <properties>
         <!-- Plugins -->
         <version.bridger>1.6.Final</version.bridger>
-        <version.module-info>2.1</version.module-info>
         <version.sundrio>0.200.0</version.sundrio>
         <version.jandex-maven-plugin>1.2.3</version.jandex-maven-plugin>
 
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -160,9 +160,6 @@
                 -->
                 <plugin>
                     <artifactId>maven-javadoc-plugin</artifactId>
-                    <configuration>
-                        <legacyMode>true</legacyMode>
-                    </configuration>
                 </plugin>
                 <plugin>
```
radcortez, line 164: Do we still use the old plugin?

dmlloyd: After this PR, all the modules are javadoc'd as JPMS modules so we don't need `legacyMode` anymore. We do still use the javadoc plugin though if that's what you're asking?

radcortez: No, the `<groupId>io.github.dmlloyd.module-info</groupId>` :)

dmlloyd: Ah, no it should be removed now.

dmlloyd: Fixed

radcortez (approved):

PR #376, opened 2024-11-19
Fix javadoc issue by swapping task order

dmlloyd:
Also quiet the build a little


dmlloyd:
I updated it to build javadocs during the normal build. It only adds a few seconds, and this way CI will break if the docs are broken.

dmlloyd:
This also reveals some javadoc issues that need fixing.

radcortez (approved):

PR #379, opened 2024-11-25
Use `JarFile.OPEN_READ` instead of `ZipFile`

gastaldi:
/cc @dmlloyd

### DIFF
--- a/resource/src/main/java/io/smallrye/common/resource/JarFileResourceLoader.java
+++ b/resource/src/main/java/io/smallrye/common/resource/JarFileResourceLoader.java
@@ -7,7 +7,6 @@ import java.nio.file.Path;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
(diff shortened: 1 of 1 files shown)


dmlloyd:
I see why you did this change, however the `OPEN_READ` constant is actually declared on `ZipFile` and is only "inherited" (in the way of `static` members) onto the subclass. Even `JarFile` itself references `ZipFile.OPEN_READ`, so I would say that it is correct as-is. But if you have a strong argument the other way, I'll hear it.

gastaldi:
Well, `JarFile.OPEN_READ` is used in https://github.com/smallrye/smallrye-common/pull/379/files#diff-1c013c4a7409c4eba326aba385c6b75c3f3786b9c1534f70096714291dd84100R43 so I thought it would be nice to follow :)

dmlloyd:
That's fair, they should be consistent. I wish I would have done `ZipFile` on the other place too, but I guess consistency is more important. :-)

dmlloyd (approved):

PR #385, opened 2024-12-03
Binary search API

dmlloyd:
Will update the summary/description once things are more finalized.

Design (short version):

* `BinarySearch.intRange().find(0, 100, n -> n >= 35)` - find the first integer in range `[0, 100)` which satisfies the predicate `n >= 35` (in this case, yields 35 after about 7 steps)
* `BinarySearch.longRange().find(1000L, 1000000L, n -> n >= 2048L)` - find the first long integer in range `[1000L, 1000000L)` which satisfies the predicate `n >= 2048L`

We also have `BinarySearch.intRange().unsigned()` which treats the range as unsigned integers (`0..0xFFFF_FFFF`), and `BinarySearch.longRange().unsigned()` which is equivalent for `long`.

There are also variations which allow you to provide your own midpoint function.

There are helpers to search ranges of things, for example, given:

```java
public record Item(String name, String address) {}
```

We can search a `List<Item> list` for the given `name` using a helper like this:

`int idx = BinarySearch.intRange().findFirst(list, "Fred", 0, list.size(), (l, i) -> l.get(i).name())`

Then `idx` will either contain the exact match, or the point into which the exact match would go.

See https://bugs.openjdk.org/browse/JDK-8326330 for the original inspiration of this flexible implementation.

I'm not super happy with the API ergonomics (the factory methods). But it seemed better than having one `BinarySearch` class with 50 or more methods called `find`. And the factory methods are all monomorphic and don't actually allocate anything, so the perf impact is expected to be minimal as these can be trivially inlined. But I'm open to other ideas in terms of API ergonomics and naming of classes.

### DIFF
--- a/function/src/main/java/io/smallrye/common/function/ExceptionObjIntFunction.java
+++ b/function/src/main/java/io/smallrye/common/function/ExceptionObjIntFunction.java
@@ -1,0 +1,33 @@
+package io.smallrye.common.function;
+
+import java.util.Objects;
+
+/**
+ * A function which operates on an object and an integer, yielding an object.
+ *
+ * @param <T> the first argument type
+ * @param <R> the result type
+ * @param <E> the exception type
+ */
+public interface ExceptionObjIntFunction<T, R, E extends Exception> {
+    /**
+     * Applies this function to the given arguments.
+     *
+     * @param arg1 the first argument
+     * @param arg2 the second argument
+     * @return the function result
+     * @throws E if an exception occurs
+     */
+    R apply(T arg1, int arg2) throws E;
+
+    /**
+     * {@return a function that applies the given function to the result of this function}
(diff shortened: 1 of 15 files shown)


dmlloyd:
OK I'm pretty happy with the API now. Just going to add some more tests before marking this as ready to go.

dmlloyd:
OK everything is all set. The artifact name is `smallrye-common-search` and the module name is `io.smallrye.common.search`. We could potentially add support for other kinds of search algorithms in the future to this module.

dmlloyd:
Rebased to conform to new javadoc validation rules.

radcortez (approved):

radcortez:
I had a look; there is a lot of code, but high level it looks fine :)

PR #389, opened 2025-01-20
Make CPU cache interrogation an opt-in feature

geoand:
This feature has a measurable startup performance
and for most modern CPUs, the cache line size is
64 anyway, so it doesn't make sense burning cycles to figure it out just for the odd case were the value is different

### DIFF
--- a/cpu/src/main/java/io/smallrye/common/cpu/CacheInfo.java
+++ b/cpu/src/main/java/io/smallrye/common/cpu/CacheInfo.java
@@ -78,82 +78,93 @@ public final class CacheInfo {
     }
 
     static {
-        cacheLevels = doPrivileged((PrivilegedAction<CacheLevelInfo[]>) () -> {
-            try {
-                String osArch = System.getProperty("os.name", "unknown").toLowerCase(Locale.US);
-                if (osArch.contains("linux")) {
-                    // try to read /sys fs
-                    final File cpu0 = new File("/sys/devices/system/cpu/cpu0/cache");
-                    if (cpu0.exists()) {
-                        // great!
-                        final File[] files = cpu0.listFiles();
-                        if (files != null) {
-                            ArrayList<File> indexes = new ArrayList<File>();
-                            for (File file : files) {
-                                if (file.getName().startsWith("index")) {
-                                    indexes.add(file);
+        cacheLevels = doPrivileged(new PrivilegedAction<>() {
+            @Override
+            public CacheLevelInfo[] run() {
+                if (determineCacheLevel()) {
+                    try {
+                        String osArch = System.getProperty("os.name", "unknown").toLowerCase(Locale.US);
+                        if (osArch.contains("linux")) {
+                            // try to read /sys fs
+                            final File cpu0 = new File("/sys/devices/system/cpu/cpu0/cache");
+                            if (cpu0.exists()) {
+                                // great!
+                                final File[] files = cpu0.listFiles();
+                                if (files != null) {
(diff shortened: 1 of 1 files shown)


cpu/src/main/java/io/smallrye/common/cpu/CacheInfo.java
```diff
@@ -78,82 +78,90 @@ public static int getSmallestInstructionCacheLineSize() {
...
+                // all has failed
+                return new CacheLevelInfo[0];
+            });
+        } else {
+            cacheLevels = new CacheLevelInfo[0];
+        }
+    }
+
+    private static boolean determineCacheLevel() {
+        return Boolean.parseBoolean(System.getProperty("smallrye.cpu.determine-cache-level", "false"));
```
geoand, line 164: This name is far from ideal, so I'm certainly open to suggestions

radcortez: It is not going to see massive usage, so I think it is ok.

cpu/src/main/java/io/smallrye/common/cpu/CacheInfo.java
```diff
@@ -78,82 +78,90 @@ public static int getSmallestInstructionCacheLineSize() {
...
-            } catch (Throwable ignored) {
-            }
-            // all has failed
-            return new CacheLevelInfo[0];
-        });
+                // all has failed
+                return new CacheLevelInfo[0];
+            });
+        } else {
+            cacheLevels = new CacheLevelInfo[0];
```
radcortez, line 159: Should we return `64` by default?

https://github.com/JCTools/JCTools/blob/a17b56f09cc9ec7d773707578c89f3ca9c11f968/jctools-core/src/main/java/org/jctools/util/PortableJvmInfo.java#L21

geoand: I was pondering that, but I would like to hear from @dmlloyd who knows how the API was meant to be used

dmlloyd: Yes let's use 64 by default.

cpu/src/main/java/io/smallrye/common/cpu/CacheInfo.java
dmlloyd: This does have to be within `doPrivileged` unfortunately. Maybe this can be combined with the block above.

cpu/src/main/java/io/smallrye/common/cpu/CacheInfo.java
geoand: I pushed a change. Is this what you had in mind?

dmlloyd: That works. It would be marginally simpler though if the call to `determineCacheLevel` were put *inside* the existing privileged block and just short-circuits the logic.

geoand: PR updated

dmlloyd (approved):

radcortez (approved):

radcortez:
Can you sign the commit?

geoand:
Done

radcortez:
Thanks!

geoand:
Do we have a plan about making a release and including it in Quarkus?

radcortez:
I'll do a release during this week.

geoand:
🙏🏽

PR #397, opened 2025-02-18
Remove references to `SecurityManager` and `PrivilegedAction`s

### LINKED ISSUE #396: Remove all use of `SecurityManager`
`SecurityManager` was deprecated in JDK 17 with https://openjdk.org/jeps/411 and permanently disabled in JDK 24 (https://openjdk.org/jeps/486).

gastaldi:
- Fixes #396

/cc @dmlloyd


dmlloyd (changes requested):
On hold until we can work with the WildFly team on a timeline.

radcortez:
I've spoken with them recently, and they still need some time to move on.

Everything that supports Jakarta 10 needs to retain security manager support, once only Jakarta EE 11 is exclusively supported security manager support is no longer required.  EAP 8.2, by default, will be Jakarta EE 10.

The plan is to fork EAP 8.2 from WF 41 in July, after which point WF is freed from supporting EAP 8.x, so EAP 9 and WF 42 should be safe remove SM.

cescoffier:
@radcortez any more update?

radcortez:
The idea was to wait until WF 41 to remove SM, which is scheduled for July.

We can always move with this, knowing that we may have to support a version with SM included.

dmlloyd:
Let's do a major version bump when we make this change.

dmlloyd (approved):

dmlloyd:
Commits are not signed, could you please fix @gastaldi?

gastaldi:
@dmlloyd fixed

Issue #402, opened 2025-03-21
Introduce parent-child structure for duplication context


cescoffier:
We’ve encountered several scenarios where the flat structure of the existing duplicated context system is too limiting. Specifically, it becomes difficult to manage an isolated state without risk of interference or context leakage when handling sub-processing steps, such as making a REST client call as part of an HTTP request.

To address this, we propose extending the `ContextLocals` and `VertxContext` APIs to support nested duplicated contexts. This will enable a tree-like structure, where each duplicated context can reference its parent, enabling better scoping and isolation of sub-processing logic.

While not the cleanest architectural change, this approach avoids a large-scale rewrite and minimizes impact across Quarkus and SmallRye Reactive Messaging.

## Overview of the Changes

We introduce _parent-child_ relationships between duplicated contexts. Currently, duplicating a duplicated context misleadingly duplicates the root context. With this new design, we aim to preserve hierarchy, provide fine-grained control over context inheritance, and the ability to write to the local duplicated context and its parent **if** the parent is not a root context (event loop).

## New Semantics for Context Duplication

A new method will be added to `VertxContext` to support proper nested duplication:

```java
Context duplicate(Context context);
Context duplicate(); // uses current context
```

**Duplication logic:**
	 
. If the current context is null, throw an exception.
. If the current context is a root context, perform a standard duplication.
. If the current context is already duplicated:	

     .	Create a new duplicated context.
     .  Copy the local map from the current context.
     . Set a special _parent reference to the current duplicated context.

This allows nesting duplicated contexts to represent sub-processing flows.

## Additions to ContextLocals

We extend the API to support parent context introspection and interaction:

* `Context getParentContext()`: Returns the parent of the current duplicated context, if any. Returns `null` if there is no `parent` context.
* `boolean putInParent(String key, Object value)`: Writes a value to the parent context’s map. Throws an exception if no parent exists. This behaves like `putIfAbsent` — returns false if the key already exists in the parent. This prevents multiple concurrent writes to a single key.
* `Object getFromParent(String key)` - Retrieves a value from the parent context. Throws an exception if no parent exists. This method should be rarely used, as nested duplicated contexts inherit the current snapshot of values from the parent. This is intended for scenarios where the most up-to-date value is required.

## Optional Additions

To improve lifecycle management, we can optionally introduce:

* `void close()`: Marks the current duplicated context as read-only. Any further mutations (via put, etc.) will result in exceptions.
[...]

PR #403, opened 2025-03-21
Parent-child duplicated context primitives

jponge:
See #402


jponge:
Looking at iterations for #402

PR #417, opened 2025-05-30
Added SuppressForbidden annotation.

eniuane:


annotation/src/main/java/io/smallrye/common/annotation/SuppressForbidden.java
```diff
@@ -0,0 +1,15 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Used to suppress forbidden-apis errors inside the element it's added to.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
+public @interface SuppressForbidden {
+    String reason();
```
Pankraz76, line 14:
```suggestion
	@AliasFor("reason")
	String[] value() default {};

	@AliasFor("value")
	String[] reason() default {};
```
might overload like done in org.springframework.context.annotation.Bean.

eniuane:
Exactly it’s regarding a PR discussion in quarkus, with @Ladicek and @mkouba

dmlloyd (changes requested):
Sorry, that explanation explains nothing. What is the function of this annotation? What is a "forbidden" API? At the very least, the documentation should explain exactly what it's for *without* assuming some prior context.

eniuane:
@dmlloyd , already updated the JavaDoc, let me know what you think and how can I improve it even more.

Ladicek:
@dmlloyd This is an annotation we currently have on 3 places in Quarkus that's used together with https://github.com/policeman-tools/forbidden-apis/. That's a Maven plugin that analyzes bytecode to find usages of forbidden APIs, and this annotation suppresses the check for the annotated element.

annotation/src/main/java/io/smallrye/common/annotation/SuppressForbidden.java
eniuane: If possible I would prefer to maintain `reason` just for us to be able to change it in the future if needed.

eniuane:
Removed @snipped from the JavaDoc, to be compatible with Java 17.

annotation/src/main/java/io/smallrye/common/annotation/SuppressForbidden.java
Pankraz76: > bad idea

the idea is not enforcing some **random param**, as the topic is kind clear, you just mention it, there wont be any more very likely.

Considering demo PR, its about avoiding **boilerplate** and go with concise API, just like spring does.
Providing the param for devs who like more overhead and verbose code can do so, just like in spring. 

- https://github.com/quarkusio/quarkus/pull/48167
- https://www.baeldung.com/spring-aliasfor-annotation#aliasfor

As the property seems not used anyway its pure sugar.

eniuane:
@dmlloyd I just finished  all the changes, let me know if everything is OK.

annotation/src/main/java/io/smallrye/common/annotation/SuppressForbidden.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Annotation that specifies a suppression inside the element it's added to.
+ * <p>
+ * Used to configure maven plugin {@code de.thetaphi.forbiddenapis} that analyzes bytecode
+ * to find usages of forbidden APIs
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
+public @interface SuppressForbidden {
+    String reason();
```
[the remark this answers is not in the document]
Pankraz76: Its worth doing well. 
Enforcing the only parameter might be clunky as resulting in boilerplate.

Ladicek (approved):
LGTM in its current form.

I admire the passion, @Pankraz76, but there's exactly _zero_ need to make this more complex than it needs to be. This is an internal annotation, not some kind of public API.

Pankraz76:
> more complex

less parameter = less complex.
enforcing parameter = more complex.

imho.

> public API

technically right, its the interface between dev and system, therefore kind of api imho.

annotation/src/main/java/io/smallrye/common/annotation/SuppressForbidden.java
```diff
@@ -0,0 +1,22 @@
...
+ * should be allowed even if there is a policy forbidding it.
+ * <p>
+ * A valid reason must be given for this kind of exception; therefore,
+ * the {@link #reason() reason} element is required.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.LOCAL_VARIABLE, ElementType.METHOD, ElementType.TYPE })
+public @interface SuppressForbidden {
+    /**
+     * {@return the reason for allowing a forbidden API in the annotated element}
```
Pankraz76, line 19:
```suggestion
     * {@return the reason for allowing a forbidden API in the annotated element.}
```

Pankraz76: not needed, only polish.

Pankraz76 (approved):
well done.

annotation/src/main/java/io/smallrye/common/annotation/SuppressForbidden.java
Ladicek: This would actually be wrong, the period is added automatically in this case.

Pankraz76: ah nice feature.

dmlloyd (approved):

gsmet:
@dmlloyd could this be backported to `2.x`? I'd like to make progress on ForbiddenAPIs usage and would love to have this one around.

PR #418, opened 2025-06-04
Introduce process API

dmlloyd:
As promised in https://github.com/quarkusio/quarkus/discussions/48223, here's the draft Process API.

The tests (so far) pass. I want to add more tests and try replacing some usages in Quarkus itself before I will call this ready to merge, but feel free to review now.


process/src/main/java/io/smallrye/common/process/Gatherer.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+/**
+ *
+ */
+final class Gatherer implements Consumer<Object> {
```
geoand, line 13: Can we have a little Javadoc on this?

process/src/main/java/io/smallrye/common/process/AbnormalExitException.java
```diff
@@ -0,0 +1,172 @@
...
+     * @param hardTimeoutElapsed {@code true} if the hard timeout elapsed before the process exited
+     */
+    public void setHardTimeoutElapsed(final boolean hardTimeoutElapsed) {
+        this.hardTimeoutElapsed = hardTimeoutElapsed;
+    }
+
+    /**
+     * {@return the captured error output of the process execution, if any}
+     */
+    public List<String> errorOutput() {
```
maxandersen, line 121: is this stderr only ? what about stdout?

dmlloyd: We *could* capture stdout here, but doing so would mean adding "tee" support broadly. This can be done, but would require more effort. It's also possible that we could introduce this in the future.

process/src/main/java/io/smallrye/common/process/ProcessExecutionException.java
```diff
@@ -0,0 +1,49 @@
+package io.smallrye.common.process;
+
+import java.io.Serial;
+
+/**
+ * An exception in process execution.
+ */
+public class ProcessExecutionException extends RuntimeException {
```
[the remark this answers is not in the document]
dmlloyd: This is done.

PR #421, opened 2025-06-18
Enhancements for the process API

dmlloyd:
* Introduce support for the "native" charset rather than relying on `Process` to do it
* `toStringList` should have an argument which bounds the maximum string length as well as the list length
* Access control issues on JDK 24+
* Tee support for optionally capturing output in exceptions


dmlloyd:
One more coming, to introduce the tee support required in order to capture stdout in exceptions...

PR #426, opened 2025-06-23
Process API docs and a small behavior update

dmlloyd:


geoand:
I think what would be useful is to document in the Javadoc the high level problems with the JDK's `ProcessBuilder` that lead to the creation of this module

PR #427, opened 2025-06-25
Add async process start

dmlloyd:
Add an API which starts a process but returns a `CompletableFuture` instead of blocking. This can be used for tasks which start a process and expect it to run in the background without requiring those cases to start their own threads.

Note that like the rest of the process handling, this does require a dedicated thread.


process/src/main/java/io/smallrye/common/process/ProcessRunner.java
```diff
@@ -142,6 +174,18 @@ O run() {
         };
     }
 
+    private static ThreadFactory threadFactory() {
+        // todo: if (Thread.currentThread().isVirtual()) { ... }
+        return task -> new Thread(() -> {
```
gastaldi, line 179: Maybe set a name in the Thread for debugging purposes?

Issue #428, opened 2025-06-25
Add batch/powershell launching support


dmlloyd:
On Windows, powershell and (especially) batch have unusual argument handling rules that are fairly tricky to implement correctly. Therefore, at present, executing powershell and batch scripts is not allowed on Windows and will result in an exception.

Add support for these by constructing safe command lines for these cases.

For batch, since some strings are impossible to pass to batch scripts, there should be extra validation to reject these arguments immediately to avoid user confusion or (worse yet) vulnerability to injection attacks.

Note that it is technically possible to install and use powershell on other operating systems. Nevertheless, the special support should be restricted to Windows. For other operating systems, particularly POSIX-like OSes like MacOS and Linux, the `+x` file permission bit should be used with an appropriate `#!` line.

- [x] Basic support for `.bat`/`.cmd`, rejecting any characters which need quoting (#432)
- [x] Basic support for `.ps1` (#432)
- [x] Support fancy quoting for `.bat`/`.cmd` (still reject strings that are not quotable; maybe introduce a special flag which indicates that sending quoted arguments to the script is OK i.e. it uses e.g. `%~1` instead of `%1`) (#434)
- [ ] Support for argument files when the argument list is very long

/cc @maxandersen

PR #432, opened 2025-06-26
Support execution of batch and powershell scripts

dmlloyd:
This is only the minimum level of basic support. There are more things that can be done to enhance this, so it does not fully address #428 but it does partially resolve it.


gastaldi (approved):
LGTM

PR #443, opened 2025-07-07
Introduce `Exec` factory class

### LINKED ISSUE #439: different name than ProcessBuilder
discussed in #418 reusing `ProcessBuilder` name is not great.

In particular when you write code like this:

```
void main() {
    var pb = new ProcessBuilder("ls", "-l");
}
```

You get no error because ProcessBuilder is in java.lang and it is now quite tedious because the IDE will not help you import it.

How about `ProcBuilder` or `ExecBuilder` ?

dmlloyd:
Fixes #439


dmlloyd:
This isn't really necessary at this stage; the existing API is good enough.

Issue #463, opened 2025-08-19
Could we add a utility to delete the content of a whole directory?


gsmet:
I'm trying to get rid of our usage of Commons IO in Quarkus. For most of the use cases, I can just use java.nio instead but there are a few uses cases for which it would be nice to have some proper util in SmallRye Common.

A few of them I could spot in our code base, not saying all of them should be added but we know they are used:

- `org.apache.common.io.FileUtils.deleteDirectory`: delete an entire directory
- `org.apache.common.io.FileUtils.cleanDirectory`: clean the content of a directory: this one
- `org.apache.common.io.FileUtils.deleteQuietly`: delete whatever is there (file or entire directory) quietly, no exception if it fails
- `org.apache.common.io.FileUtils.copyFileToDirectory`: will copy a file to a directory, preserving the timestamp and creating the directory structure if needed
- `org.apache.common.io.FileUtils.moveDirectory`: move or copy/delete a directory when not on same fs
One is the ability to delete an entire directory: 
- `org.codehaus.plexus.util.FileUtils.copyDirectoryStructure`: we are using it once in test code

I will update the list while I investigate this.

Ideally if we could start a class with some common utilities that are missing from the JDK, that would be cool.

gastaldi:
I've prototyped an API with these methods in https://github.com/gastaldi/smallrye-common/blob/files/io/src/main/java/io/smallrye/common/io/Files.java

gastaldi:
> [@gastaldi](https://github.com/gastaldi) I wonder about the name. Because I'm pretty sure we don't want it to clash with `java.nio.file.Files`.
> 
> I don't see a lot of options apart adding a suffix but it's not ideal given prior SmallRye Common naming.

Yeah, same here. I followed the pattern we used in other classes (eg. JarFiles). I don't have a strong preference TBH, open for suggestions

dmlloyd:
I think this is all done now.

PR #482, opened 2025-09-25
Add file operations for recursive deletion/cleaning and supporting methods

dmlloyd:
This is a start to implementing a secure file utilities API as described in #463 based on @gastaldi's work and @Ladicek's dire warnings.

/cc @gsmet

### DIFF
--- a/io/src/main/java/io/smallrye/common/io/Files2.java
+++ b/io/src/main/java/io/smallrye/common/io/Files2.java
@@ -174,17 +174,17 @@ public final class Files2 {
      * @param path the path to delete (must not be {@code null})
      * @return the deletion statistics (not {@code null})
      *
-     * @see #deleteRecursivelyQuiet(Path)
+     * @see #deleteRecursivelyQuietly(Path)
      */
-    public static DeleteStats deleteRecursivelyQuietEvenIfInsecure(Path path) {
+    public static DeleteStats deleteRecursivelyQuietlyEvenIfInsecure(Path path) {
         checkNotNullParam("path", path);
         SecureDirectoryStream<Path> base = CWD_SDS;
         if (base != null) {
             // secure!
-            return deleteRecursivelyQuiet(base, path);
+            return deleteRecursivelyQuietly(base, path);
         } else {
             long[] stats = new long[DeleteStats.STATS_SIZE];
-            deleteRecursivelyQuietInsecurely(path, stats);
+            deleteRecursivelyQuietlyInsecurely(path, stats);
             return new DeleteStats(stats);
         }
     }
@@ -203,15 +203,15 @@ public final class Files2 {
      * @param ds the directory stream whose contents should be removed (must not be {@code null})
      * @return the deletion statistics (not {@code null})
      *
-     * @see #deleteRecursivelyQuiet(SecureDirectoryStream)
+     * @see #deleteRecursivelyQuietly(SecureDirectoryStream)
      */
-    public static DeleteStats deleteRecursivelyQuietEvenIfInsecure(DirectoryStream<Path> ds) {
+    public static DeleteStats deleteRecursivelyQuietlyEvenIfInsecure(DirectoryStream<Path> ds) {
         checkNotNullParam("ds", ds);
         if (ds instanceof SecureDirectoryStream<Path> sds) {
-            return deleteRecursivelyQuiet(sds);
+            return deleteRecursivelyQuietly(sds);
         } else {
             long[] stats = new long[DeleteStats.STATS_SIZE];
-            deleteRecursivelyQuietInsecurely(ds, stats);
+            deleteRecursivelyQuietlyInsecurely(ds, stats);
             return new DeleteStats(stats);
         }
     }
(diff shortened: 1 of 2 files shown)


io/src/main/java/io/smallrye/common/io/Messages.java
```diff
@@ -0,0 +1,23 @@
...
+import static java.lang.invoke.MethodHandles.lookup;
+
+import java.nio.file.FileSystem;
+import java.nio.file.Path;
+
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageBundle;
+
+@MessageBundle(projectCode = "SRCOM", length = 5)
+interface Messages {
```
gastaldi, line 12: Maybe also use it for logging?

```suggestion
interface Messages extends BasicLogger {
```

io/src/main/java/io/smallrye/common/io/Files2.java
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @see #deleteRecursively(Path)
+     */
+    public static void deleteRecursivelyEvenIfInsecure(Path path) throws IOException {
+        checkNotNullParam("path", path);
+        path = path.normalize();
+        if (!Files.exists(path)) {
+            return;
+        }
+        Path parent = path.getParent();
```
gastaldi, line 71: I think a test would be nice, but AFAIU wouldn't this delete the parent of a directory? eg. if you pass `/etc/foo` (`foo` being a directory) it will delete the `/etc` directory recursively?

dmlloyd: Tests will come next, buts they're a bit more complex for things like this so I want to be sure that we are happy with the api first.

This method opens the directory of the parent and then only recursively deletes the one file inside the parent below using the secure methods. This is because there is no option to not follow symlinks when opening a directory normally.

dmlloyd:
Just a couple of tweaks based on feedback and a little API adjustment.

One thing worth noting is that Windows (and Mac OS, until JDK 24) does not support secure directory deletion. So we'll have to consider carefully the implications of that, if we are not already.

io/src/main/java/io/smallrye/common/io/Messages.java
```diff
@@ -0,0 +1,27 @@
...
+import java.nio.file.FileSystem;
+import java.nio.file.Path;
+
+import org.jboss.logging.BasicLogger;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+@MessageLogger(projectCode = "SRCOM", length = 5)
+interface Messages extends BasicLogger {
+    Messages log = org.jboss.logging.Messages.getBundle(lookup(), Messages.class);
```
gastaldi, line 14: Shouldn't this be a Message Logger now that it extends `BasicLogger`? 

```suggestion
    Messages log = org.jboss.logging.Logger.getMessageLogger(lookup(), Messages.class, Messages.class.getPackageName());
```

dmlloyd: The lack of tests really shows, eh? :)

io/src/main/java/io/smallrye/common/io/Files2.java
```diff
@@ -0,0 +1,291 @@
...
+    public static Path currentDirectory() {
+        return CWD;
+    }
+
+    // -- private --
+
+    private static final Path CWD = Path.of(System.getProperty("user.dir", ".")).normalize().toAbsolutePath();
+
+    private static SecureDirectoryStream<Path> newSecureDirectoryStreamFollow(Path path) throws IOException {
+        DirectoryStream<Path> ds = Files.newDirectoryStream(path);
```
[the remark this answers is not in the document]
gastaldi: Hm, maybe not, it would be up to the caller to close it I believe

io/src/main/java/io/smallrye/common/io/Files2.java
```diff
@@ -238,9 +247,28 @@ public static Path currentDirectory() {
...
+     * Some operating systems or JVM versions do not support secure directories.
+     */
+    public static boolean hasSecureDirectories() {
+        return HAS_SDS;
+    }
+
     // -- private --
 
     private static final Path CWD = Path.of(System.getProperty("user.dir", ".")).normalize().toAbsolutePath();
+    private static final boolean HAS_SDS;
```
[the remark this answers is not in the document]
gastaldi: Perhaps don't store `CWD` as a static field to avoid the trouble? Not sure if it's worth the trouble for something that may not affect performance that much

io/src/test/java/io/smallrye/common/io/Files2Test.java
```diff
@@ -0,0 +1,203 @@
...
+        DeleteStats stats = Files2.deleteRecursivelyQuiet(testArea);
+        assertFalse(Files.exists(testArea));
+        assertEquals(6, stats.filesFound());
+        assertEquals(6, stats.filesRemoved());
+        assertEquals(4, stats.directoriesFound());
+        assertEquals(4, stats.directoriesRemoved());
+    }
+
+    @Test
+    public void testDeleteRecursivelyQuietEvenIfInsecure() {
```
gastaldi, line 166: You may already know, but you can also ask JUnit to create a temporary dir for your tests:


```suggestion
    public void testDeleteRecursivelyQuietEvenIfInsecure(@TempDir Path tempDir) {
```

dmlloyd: Nice, I'll try that.

dmlloyd: Ah, unfortunately that always gives an absolute path, which makes it hard to test relative path stuff. But I can still use it for absolute path tests.

gsmet:
I’ll have a look soon. I’ve been busy preparing Devoxx.

PR #486, opened 2025-10-01
Introduce secure copy, recursive copy, and move functionality

### LINKED ISSUE #484: Files extra utilities: recursive file copying and moving
Introduce secure recursive file copying and moving utilities.

dmlloyd:
Fixes #484.

This is a draft to allow review of the API and basic implementation as I write tests. Based on #482.

- [x] `copy` (SDS input)
   - [x] Tests
- [x] `copy` (Path input only)
   - [x] Tests
- [x] `copyRecursively` (SDS input)
   - [x] Tests
- [x] `copyRecursively` (Path input only)
   - [x] Tests
- [x] `move` (SDS input)
   - [x] Tests
- [x] `move` (Path input only)
   - [x] Tests
- [x] `copyEvenIfInsecure` (Path input only)
   - [x] Tests
- [x] `copyRecursivelyEvenIfInsecure` (Path input only)
   - [x] Tests
- [x] `moveEvenIfInsecure` (Path input only)
   - [x] Tests

### DIFF
--- a/io/src/main/java/io/smallrye/common/io/Files2.java
+++ b/io/src/main/java/io/smallrye/common/io/Files2.java
@@ -1,11 +1,20 @@
 package io.smallrye.common.io;
 
+import static io.smallrye.common.constraint.Assert.checkNotNullArrayParam;
 import static io.smallrye.common.constraint.Assert.checkNotNullParam;
+import static io.smallrye.common.constraint.Assert.impossibleSwitchCase;
 import static io.smallrye.common.io.Messages.log;
 
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.ByteOrder;
+import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ReadableByteChannel;
+import java.nio.channels.SeekableByteChannel;
+import java.nio.channels.WritableByteChannel;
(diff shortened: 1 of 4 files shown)


dmlloyd:
One issue I struggled a little bit with is in regards to one detail of copying.

Most copy utilities will look at your destination path, and if it's a directory, will copy the source file/directory into the target directory under the same name as the original. If it's a file that exists, it will attempt to replace that file with the copy (or fail if you don't have replacement enabled).

The JDK however behaves differently with its (strictly non-recursive) `Files.copy()` method. If the target is a non-empty directory, the copy fails (even if you specify `REPLACE_EXISTING`). If it is a file or an empty directory, and `REPLACE_EXISTING` is set, then the original file is deleted before copying.

One other little detail is that on UNIX-like systems, if your destination path ends with `/` then it's always taken to mean a directory to copy the original to, and will fail if the path is actually a file. With `java.nio.file.Path`, any trailing `/` are removed during parsing, so that behavior is not possible to implement.

The proposed API follows the JDK behavior, not because I think it's better, but because with the restriction on `Path` it would be basically impossible to be able to specify that the destination is a directory into which the file should be copied/moved. Also, I think that deviating a lot from the JDK idioms would be a source of confusion and bugs.

PR #497, opened 2025-11-06
Fix ProcessUtil.pathOfCommand(Path path) on Windows

Eng-Fouad:
Currently, when I execute the following on my Windows machine:

```
ProcessUtil.pathOfCommand(Path.of("docker"));
```

it returns `C:\Program Files\Docker\Docker\resources\bin\docker` because the file `docker` does exist, while what we are looking for is `docker.exe`:

ssss

This PR will prioritize selecting executable files with extension over executable files without extension, on Windows only.
Or should we drop looking for executable files without extension on Windows at all?


Eng-Fouad:
Pushed a signed commit.

PR #510, opened 2026-01-02
Add module annotations for `Add-Opens`, `Add-Exports`, `Enable-Native-Access`

dmlloyd:
These annotations may be consumed by tooling or by the upcoming Quarkus module loader implementation.

/cc @gastaldi @Sanne @cescoffier


gastaldi (dismissed):
LGTM

gastaldi (dismissed):

Ladicek:
I think it would be good if the javadoc explicitly said that the annotations themselves are _not_ enough and they only work in a cooperating runtime. LGTM otherwise.

gastaldi (approved):

PR #513, opened 2026-01-15
Upgrade to Vert.x 5

jponge:
This is part of the Quarkus to Vert.x 5 effort: https://github.com/quarkusio/quarkus/issues/51959


jponge:
How about if we split into `vertx4-context` and `vertx5-context`?

jponge:
I'll re-sync with the upcoming 4.x and 5.x releases, then make the PR ready.

dmlloyd (approved):

vertx5-context/src/main/java/io/smallrye/common/vertx/VertxContext.java
```diff
@@ -0,0 +1,293 @@
...
+     * The ContextLocal used to store the parent context in a nested context.
+     */
+    public static final ContextLocal<Context> PARENT_CONTEXT_LOCAL = ContextLocal.registerLocal(Context.class);
+
+    /**
+     * The ContextLocal used to store arbitrary key-value data in a context.
+     */
+    @SuppressWarnings("unchecked")
+    public static final ContextLocal<ConcurrentHashMap<String, Object>> DATA_MAP_LOCAL = (ContextLocal<ConcurrentHashMap<String, Object>>) (ContextLocal<?>) ContextLocal
+            .registerLocal(ConcurrentHashMap.class, ConcurrentHashMap::new);
```
cescoffier: If I understand correctly, this creates the map on the first access. Any possibility to create it eagerly?

vertx5-context/src/main/java/io/smallrye/common/vertx/internal/ContextLocalLoader.java
```diff
@@ -0,0 +1,39 @@
...
+ * <p>
+ * This is necessary because {@link VertxContext} contains {@code ContextLocal} field registrations
+ * that must be initialized before any Vert.x instance is created.
+ * </p>
+ * <p>
+ * The Vert.x ServiceLoader mechanism loads this provider during Vert.x initialization,
+ * which triggers the loading of {@link VertxContext} and its static {@code ContextLocal} fields.
+ * </p>
+ */
+public class ContextLocalLoader implements VertxServiceProvider {
```
cescoffier: When is this provider initialized? Can we do it explicitly?

jponge: AFAIK it gets initialized by Vert.x (ping @tsegismont to confirm)

vertx5-context/src/main/java/io/smallrye/common/vertx/VertxContext.java
jponge: `ConcurrentHashMap::new` is a duplicator function (aka `ConcurrentHashMap(map)`), not a supplier. This is called when you `ctx.duplicate(copy=true)`.

vertx5-context/src/main/java/io/smallrye/common/vertx/internal/ContextLocalLoader.java
tsegismont: It's initialized automatically by Vert.x if not provided by the user. But @vietj and @gsmet collaborated recently so that Quarkus can provide the list of service providers to the vert.x builder https://github.com/quarkusio/quarkus/pull/52632/changes#diff-99bb6be11b1e4b85ac938b26fbadf0ff1d58e23d88ab481ea5882df75ab3c18eR255

PR #524, opened 2026-03-04
Use SmallRye FFM to reliably get host name

dmlloyd:
This is a quick example of how SmallRye FFM can be used. In this case, we have always relied on complicated heuristics to determine the system host name because Java has no API to access this information. But with SmallRye FFM, we can just make the native call directly if the JVM allows it.

This should not necessarily be considered for merging, at least until we have solutions for both Quarkus and WildFly to register native access. But eventually, doing something like this could replace most of the heuristics that we currently employ (like probing environment variables).

Note that while this is a large change, much of the change consists of things that should be moved to a parent POM: working around impsorter bugs, establishing common versions. Also, there are some legitimately missing tests in the `net` module relating to host name (it's hard to really test in general, but we can at least make sure we don't explode).

Edit: the Windows build doesn't presently work; this may be a problem on the SmallRye FFM side, but it can be ignored for now.

### DIFF
--- a/net/pom.xml
+++ b/net/pom.xml
@@ -15,7 +15,16 @@
     <dependencies>
         <dependency>
             <groupId>io.smallrye.common</groupId>
+            <artifactId>smallrye-common-annotation</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.smallrye.common</groupId>
             <artifactId>smallrye-common-constraint</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.smallrye.common</groupId>
+            <artifactId>smallrye-common-os</artifactId>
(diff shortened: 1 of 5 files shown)


dmlloyd:
Well, I've spent far too much effort trying to tell what the deal is with the Windows linking (which seems to work sometimes and not other times). Will have to follow up at another time once I have access to a Windows system again.

dmlloyd:
Hey, it works now. Cool.

dmlloyd:
Updated to use the new `@Lib` annotation.

dmlloyd:
Updated to reflect security manager removal. Also I realized that we wouldn't need the substitutions here anymore either with this change, so I removed that as well; a net reduction of around 45 lines of code in total compared to before.

PR #530, opened 2026-03-06
Alternative strategy for async process start and daemon processes

### LINKED ISSUE #585: run/runAsync without blocking jvm shutdown
currently ProcessBuilder blocks jvm shutdown if processes have not finished at jvm shutdown:

```
///usr/bin/env jbang "$0" "$@" ; exit $?
//JAVA 21+
//DEPS io.smallrye.common:smallrye-common-process:2.20.0

import static io.smallrye.common.process.ProcessBuilder.*;
public class SmallRyeProcessShutdown {
    public static void main(String[] args) {
        var seconds = args.length == 0 ? 3 : Integer.parseInt(args[0]);
        newBuilder("sleep", Integer.toString(seconds))
                .whileRunning(process -> System.out.println("child pid: " + process.pid()))
                .runAsync();

        System.out.println("main() returning immediately; child sleeps for " + seconds + "s");
    }
}
```

any reason why?

forces me to use other API's to kick off background jobs.

Thats sad :)

### LINKED ISSUE #588: be able to "leak" a process
follow up on #585  - would like to be able to launch processes that does not die/gets closed when the main process exits.

dmlloyd:
This strategy relies on making our waitable process handle also be closeable, which is arguably better than having CF.

Also, add support for "daemon" processes which are specifically designed to outlive the parent JVM.


dmlloyd:
WDYT @maxandersen?

process/src/main/java/io/smallrye/common/process/PipelineBuilder.java
```diff
@@ -82,6 +82,15 @@ public sealed interface PipelineBuilder<O> permits PipelineBuilder.Error, Pipeli
...
     PipelineBuilder<O> modifyEnvironment(Consumer<Map<String, String>> action);
 
+    /**
+     * Configure whether the process is a daemon process.
+     * A daemon process is allowed to outlive the parent JVM.
+     *
+     * @param daemon {@code true} to make the process a daemon, {@code false} otherwise
+     * @return this builder
+     */
+    PipelineBuilder<O> daemon(boolean daemon);
```
maxandersen: maybe also .daemon() as short cut for .deamon(true) ?

maxandersen:
look good :)

process/src/main/java/io/smallrye/common/process/PipelineBuilder.java
dmlloyd: I just got rid of the boolean parameter altogether.

PR #540, opened 2026-04-08
Reworked Vert.x 5 ContextLocal lifecycle hooks

jponge:
This should ensure that the 2 ContextLocal fields in VertxContext
always get picked up by any VertxServiceProvider handler.

Also bumped to Vert.x 5.0.10.


jponge:
@cescoffier for review

jponge:
Part of https://github.com/quarkusio/quarkus/issues/51959

cescoffier (approved):

jponge:
@dmlloyd Would you be able to make a patch release from this? 🙏

dmlloyd:
Normally I wouldn't do a patch release when the API surface has changed, but in this case it is adding just a nested class so I think it's OK. I'll get this figured out this morning.

jponge:
Thanks @dmlloyd

Issue #542, opened 2026-04-10
Vert.x 5 Context Local API should allow selecting the context


cescoffier:
Having an API like this:
```
ContextLocals.on(context).put(key.toString(), value);
```

would be really useful.

jponge:
@cescoffier is this still relevant with `VertxContext`?

jponge:
Ok I see the issue now, I see why some sugar API would help 🤣

PR #544, opened 2026-04-20
Context binding overloads in ContextLocals

jponge:
This facilitates usages of the API where we get the current context before we try to access the data.

### DIFF
--- a/vertx5-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
+++ b/vertx5-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
@@ -50,7 +50,6 @@ public class ContextLocals {
      * @param <T> the expected type of the associated value
      * @return an optional containing the associated value if any, empty otherwise.
      */
-    @SuppressWarnings("unchecked")
     public static <T> Optional<T> get(String key) {
         return get(Vertx.currentContext(), key);
     }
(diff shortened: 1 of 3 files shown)


jponge:
Fixes #542

vertx5-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
```diff
@@ -154,18 +241,33 @@ public static <T> Optional<T> getFromParent(String key) {
     /**
      * Gets the value associated with the given key from the parent context.
      * <p>
+     * If the parent context is not set, it returns an empty optional.
+     *
+     * @param key the key, must not be {@code null}
+     * @param <T> the expected type of the associated value
+     * @return an optional containing the associated value if any, empty otherwise.
+     */
+    @SuppressWarnings("unchecked")
```
cescoffier: is it required?

cescoffier (dismissed):

vertx5-context/src/main/java/io/smallrye/common/vertx/ContextLocals.java
jponge: You are right, there are also a few other `unchecked` that are redundant. Let me clean those.

cescoffier (approved):

PR #556, opened 2026-05-21
Add `crc32()` method to `Resource`

dmlloyd:


dmlloyd:
I'm not actually sure I want to bother with this one. Originally it was needed for the upcoming `smallrye-common-archive` work but in the end I found that it wasn't all that useful.

PR #562, opened 2026-06-01
Deprecate JAR utilities

dmlloyd:
These were only useful when the library could run on Java 8.

### DIFF
--- a/io/src/main/java/io/smallrye/common/io/jar/JarFiles.java
+++ b/io/src/main/java/io/smallrye/common/io/jar/JarFiles.java
@@ -8,6 +8,7 @@ import java.util.zip.ZipFile;
 /**
  * Java 9+ variant of a JDK-specific class for working with {@code JarFile}s.
  */
+@Deprecated(since = "2.19", forRemoval = true)
 public class JarFiles {
     private JarFiles() {
     }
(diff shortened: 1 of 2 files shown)


io/src/main/java/io/smallrye/common/io/jar/JarFiles.java
```diff
@@ -8,6 +8,7 @@
 /**
  * Java 9+ variant of a JDK-specific class for working with {@code JarFile}s.
  */
+@Deprecated(since = "2.19", forRemoval = true)
```
Ladicek, line 11: Note that this class is still useful, because the `create()` methods enables versioned reads, unlike `JarFile` constructors. When removed, all callers would have to use the full-blown `JarFile` ctor, which is ... possible, but not nice.

dmlloyd: It's not particularly onerous. But at any rate my thought is that we'll switch to using the upcoming SmallRye Common archive API before it would matter either way.

dmlloyd: Is that OK? I'd like to get this merged if possible.

Ladicek: Oh don't feel blocked by this comment. I was just making a point. It is quite possible I'm the first and only user of this :laughing:

PR #566, opened 2026-06-04
Introduce new `Archive` APIs to `smallrye-common-io`

dmlloyd:
* Supplies an `Archive` API for accessing ZIP/JAR archives
* Changes `JarFileResourceLoader` to use the new `Archive` API
* Introduces an `ArchiveBuilder` API for creating archives
* Updates `BufferedFile` to support nested file creation to support the above

The `Archive` and `ArchiveBuilder` APIs support nested archives.

On the read side, `Archive` maps the entire archive into memory. This means that classes can be defined by grabbing slices out of the buffer (supported by the `Resource` API). It also means that nested archives are also in the same mapping, meaning that they can be accessed in the same way. The only Java heap overhead is an index to locate the CDEs for each entry, and the memory segment itself; the index is sorted lexicographically by Unicode code point for fast lookups. On Java 22 or later, archives of >2GB can be loaded.

On the write side, `ArchiveBuilder` can write an archive in essentially one pass (unlike `JarOutputStream` which requires information like size and CRC to be supplied before writing the entry). This is done using random-access writes, so this cannot be used to (for example) stream an archive creation across a socket or anything like that (use `JarOutputStream` in this case). The archive builder also supports nested archive creation, which should be quite useful for building uber-JARs of various kinds. The API uses `Closeable` in a sane way for each entry as well.


io/src/main/java/io/smallrye/common/io/archive/Archive.java
```diff
@@ -0,0 +1,446 @@
...
+import java.util.zip.Inflater;
+
+import io.smallrye.common.annotation.Experimental;
+import io.smallrye.common.search.BinarySearch;
+
+/**
+ * A ZIP archive.
+ */
+@Experimental("Incubating archive API")
+public final class Archive {
```
gastaldi, line 28: How about naming it it `ZipArchive`? Thinking of other Archives that could be supported in future

dmlloyd: I did think about that for a bit. But my conclusion was that in Java, there's only one kind of archive, really. This appears in Maven too for example, where you just have the Archiver (and its config which is under `<archive>`). If we ever want to support other kinds then we could start introducing e.g. `TarArchive` or whatever.

resource/src/main/java/io/smallrye/common/resource/JarFileResourceLoader.java
```diff
@@ -34,79 +43,194 @@ public JarFileResourceLoader(final Path jarPath) throws IOException {
...
+         *
+         * @param resource the resource of the JAR file (must not be {@code null})
+         * @throws IOException if opening the JAR file fails for some reason
+         */
+        public ClassicImpl(final Resource resource) throws IOException {
+            if (resource instanceof PathResource pr && pr.hasFile()) {
+                // avoid using a temp file, if possible
+                jarFile = new JarFile(pr.file(), true, JarFile.OPEN_READ, JarFile.runtimeVersion());
+            } else {
+                tempFile = Files.createTempFile("srcr-tmp-", ".jar");
```
gastaldi: Do we need this snippet in here (or someplace else?) 

https://github.com/quarkusio/quarkus/commit/5560de4c1d6d03cee07c7ba33fbf3380e0648cbe#diff-cf1eb8eb33eeda443ab46eacf3c7fd5816ffc68b2ce1f12ae3b6e6df66deb496R41-R69

dmlloyd: This is the existing code so I didn't make any changes to it. I will soon delete the old code and have it use the new builder only.

dmlloyd:
Any other thoughts? @smallrye/common

PR #572, opened 2026-06-29
Add beginDispatch/endDispatch to VertxContext

gastaldi:
## Summary

- Add `VertxContext.beginDispatch(Context)` and `VertxContext.endDispatch(Context, Context)` static methods to both `vertx4-context` and `vertx5-context` modules
- These methods wrap `ContextInternal.beginDispatch()` / `endDispatch()` so that consumers can manage the thread-local context dispatch lifecycle without depending on the Vert.x internal API

## Motivation

The whole point of `VertxContext` is to shield consumers from `ContextInternal` — but until now it only covered context *creation* (duplicate, nested, etc.), not context *dispatch*. This forced projects like [quarkus-ironjacamar](https://github.com/quarkiverse/quarkus-ironjacamar) to import `ContextInternal` directly for `beginDispatch()` / `endDispatch()` calls (see [`DuplicatedContextMessageEndpoint`](https://github.com/quarkiverse/quarkus-ironjacamar/blob/main/runtime/src/main/java/io/quarkiverse/ironjacamar/runtime/endpoint/DuplicatedContextMessageEndpoint.java)).

The JCA message endpoint delivery lifecycle splits begin/end across separate method calls (`beforeDelivery` → message delivery → `afterDelivery`), so the higher-level `Context.dispatch(Runnable)` API cannot be used. These new methods fill that gap.

## Test plan

- [x] Added `testBeginAndEndDispatch` to both `vertx4-context` and `vertx5-context` test suites
- [x] Tests verify that `beginDispatch` sets the thread-local context and returns the previous one, and `endDispatch` restores it
- [x] `mvn test -pl vertx4-context,vertx5-context` passes

### DIFF
--- a/vertx4-context/src/main/java/io/smallrye/common/vertx/VertxContext.java
+++ b/vertx4-context/src/main/java/io/smallrye/common/vertx/VertxContext.java
@@ -277,4 +277,54 @@ public class VertxContext {
         return newNestedContext(Vertx.currentContext());
     }
 
+    /**
+     * Begins the execution of a task on the given context.
+     * <p>
+     * This method makes the given context the current thread-local context, so that subsequent calls to
+     * {@link Vertx#currentContext()} return this context. It returns the previous context (if any), which
+     * <strong>must</strong> be passed to {@link #endDispatch(Context, Context)} when the task is complete.
+     * </p>
+     * <p>
+     * This is a low-level API intended for situations where the begin and end of a dispatch cannot be wrapped
+     * in a single {@code Runnable} (e.g., JCA message endpoint delivery where {@code beforeDelivery} and
+     * {@code afterDelivery} are separate calls). Prefer the Vert.x {@code Context#dispatch} methods when possible.
+     * </p>
+     *
+     * @param context the context to begin dispatching on, must not be {@code null}
+     * @return the previous context that was active on this thread, or {@code null} if there was none
+     * @see #endDispatch(Context, Context)
+     */
+    public static @Nullable Context beginDispatch(Context context) {
+        ContextInternal actual = Assert.checkNotNullParam("context", (ContextInternal) context);
+        return actual.beginDispatch();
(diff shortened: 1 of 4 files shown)


Ladicek:
> The whole point of `VertxContext` is to shield consumers from `ContextInternal`

Is it? In my eyes, the whole point of `VertxContext` is to provide utilities for common operations that are not directly available on `Context` or `ContextInternal`. (I'm ignoring the whole "nested context" thing, because that should have never existed.)

There's nothing wrong with using `ContextInternal`, in my eyes.

cescoffier:
I agree with @Ladicek, I don't think this is bringing much, except exposing low-level APIs which I would not recommend using, except if you know what you are doing.

gastaldi:
Well, it's explicitly said in the class' Javadoc:

> The rationale for this class is to avoid end-users to use the Vert.x internal API.

About @cescoffier's comment, I agree, but I don't see how I can use the recommended `Context.dispatch` if the spec (eg. MessageEndpoint in JCA) splits into separate methods (same as we do in other core extensions).

dmlloyd:
What was the resolution on this one?

gastaldi:
That would be interesting to avoid having to maintain two separate codebases in quarkus-ironjacamar when consuming the Vert.x API, but no big deal if it's rejected.

dmlloyd:
I'll let @cescoffier make the final decision on this I guess.

cescoffier:
I still prefer not exposing them. If we expose them, it's not just iron jacamar but every single user that will be exposed to dangerous methods (because if you don't know what you are doing, it can be really bad - security and safety bad). We expose this API to application users, so better not add anything dangerous.

cescoffier (changes requested):
See my last comment - I would not expose these methods

gastaldi:
Ok, that settles then. Closing

PR #575, opened 2026-06-30
Add `serial` subproject

dmlloyd:
Implements an API to introspect serializable objects into a description, and also reconstitute serializable objects from a description.

This is in draft until smallrye/smallrye-parent#673 is merged.


dmlloyd:
Moving to its own repo.

PR #591, opened 2026-08-28
Expression PRESERVE_UNKNOWN_ESCAPES flag

radcortez:
Adds `Expression.Flag.PRESERVE_UNKNOWN_ESCAPES`. 

When combined with `ESCAPES`, an unrecognized escape (e.g. `\,`) is emitted verbatim with its leading backslash, instead of dropping the backslash (`LENIENT_SYNTAX`) or throwing. Recognized escapes (`\\`, `\$`, `\n`, `\r`, `\t`, `\b`, `\f`) are unaffected, and the flag is a no-op unless `ESCAPES` is set.

This would allow us to support the comma escape in values that represent arrays:
https://github.com/smallrye/smallrye-common/pull/344#issuecomment-2356616925

In `SmallRyeConfig`, we would then enable `NO_$$`, `ESCAPES` and `PRESERVE_UNKNOWN_ESCAPES` to fix:
- https://github.com/smallrye/smallrye-config/issues/1219
- https://github.com/smallrye/smallrye-config/issues/746

The only breaking change would be that `$$` no longer acts as an escape for expressions.

### DIFF
--- a/expression/src/main/java/io/smallrye/common/expression/Expression.java
+++ b/expression/src/main/java/io/smallrye/common/expression/Expression.java
@@ -609,6 +609,11 @@ public final class Expression {
                                     break;
                                 }
                                 default: {
+                                    if (flags.contains(Flag.PRESERVE_UNKNOWN_ESCAPES)) {
+                                        // keep the backslash and the following character verbatim, e.g. "\," -> "\,".
+                                        // 'start' still points at the backslash, so just keep accumulating.
+                                        continue;
+                                    }
                                     if (flags.contains(Flag.LENIENT_SYNTAX)) {
                                         // TP 40
                                         // just append the literal character after the \, whatever it was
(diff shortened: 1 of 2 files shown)


radcortez:
@dmlloyd let me know if this is ok to keep the compatibility with config and comma-separated values in arrays, or if it should also be a breaking change.

dmlloyd (changes requested):
My concern with this mode is that it introduces a potential ambiguity. If I give `\,` that would get passed through. If I give `\\,`, that is passed through as `\,`, so it now becomes impossible to have a list item that ends in `\\` - an even narrower and more confusing edge case than the one being solved.

This is why it is not recommended to have uneven escape schemes, and relatedly, it is why it is very very hard to spread non-layered lexicographic rules across layered parsers.

Escaping must remain uniform. This means that if the user wants to escape a comma for properties lists, they just have to do it the ugly (but syntactically sound) way.

PR #598, opened 2026-09-16
Switch to Maveniverse BOM builder

### LINKED ISSUE #597: Convert from sundrio to `eu.maveniverse.maven.plugins:bom-builder3` for BOM creation
The `eu.maveniverse.maven.plugins:bom-builder3` plugin is theoretically better-supported; sundrio no longer has a maintainer.

dmlloyd:
Fixes #597.

We should potentially use this as a template for other SmallRye projects as well.


dmlloyd:
See also smallrye/smallrye-parent#688.

PR #599, opened 2026-09-16
Index and UTF-8 optimizations

dmlloyd:
After some measurements, I believe this change may drastically improve smallrye/smallrye-modules#39.

/cc @gsmet


dmlloyd:
Fixed another (potential) issue: megamorphic dispatch on index implementations. The savings of the more compact index implementations was really, really small in terms of memory footprint, whereas the dispatch improvement could be significant across thousands of calls.

gsmet:
Went from 4.84% to 2.69% of the samples for `Archive.of()` in my testing.

